PR #28, opened 2019-10-09
Meepctl kibana configuration fixes

pastorsx:
NA-695 Meepctl cannot use os.Hostname()

On some systems (ex. VM), this command does not return a value that can be resolved. This command is currently used in meepctl config kibana command and should be replaced. The verbose was used for all commands except the first curl command which is what is failing when hostname cannot be resolved or kibana cannot be reached. The proper call to show the output were added.

Tested using valid Kibana pod (success path), not reachable host (failure path) and No Kibana pod (failure path) using verbose and not verbose. Results as expected following the fix.

### DIFF
--- a/go-apps/meepctl/utils/kibana.go
+++ b/go-apps/meepctl/utils/kibana.go
@@ -112,9 +112,9 @@ func isKibanaReady(cobraCmd *cobra.Command) bool {
 }
 
 func uploadDefaultIndex(indexId string, cobraCmd *cobra.Command) error {
-	kibanaHost, _ := os.Hostname()
+	kibanaHost := viper.GetString("node.ip")
 	cmd := exec.Command("curl", "-vX", "POST", "http://"+kibanaHost+":32003/api/kibana/settings/defaultIndex", "-H", "Content-Type: application/json", "-H", "kbn-xsrf: true", "-d", "{\"value\": \""+indexId+"\"}")
-	out, err := cmd.CombinedOutput()
+	out, err := ExecuteCmd(cmd, cobraCmd)
 	if err != nil {
 		err = errors.New("Error sending a curl command")
 	} else {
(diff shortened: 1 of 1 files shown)


go-apps/meepctl/utils/kibana.go
```diff
@@ -112,9 +117,9 @@ func isKibanaReady(cobraCmd *cobra.Command) bool {
 }
```
roymx: This error is never reported.
Try to adopt the 
if err != nil {
do something
}
else do something else 

structure from now on

roymx: Error in isKibanaReady is never reported.
Try to adopt the
if err != nil {
do something
}
else do something else

structure from now on

pastorsx: Error is not reported because it is only used to trigger a Boolean response. The error already got processed in the function uploadDefaultIndex called by the isKibanaReady function. That was my rationale for not duplicating reporting the error.

go-apps/meepctl/utils/kibana.go
```diff
@@ -112,9 +117,9 @@ func isKibanaReady(cobraCmd *cobra.Command) bool {
...
 
 func uploadDefaultIndex(indexId string, cobraCmd *cobra.Command) error {
```
roymx: Since the error returned by this function is not used, make it return true or false to indicate success or failure.
This will prevent creating error objects for nothing.

go-apps/meepctl/utils/kibana.go
roymx: OK - then return type of uploadDefaultIndex needs to be boolean since the error is useless

pastorsx: fixed as part of my latest commit

go-apps/meepctl/utils/kibana.go
pastorsx: fixed as part of my latest commit

PR #30, opened 2019-10-16
os.hostname() should not be used in meepctl tool

pastorsx:
Issues fixed:

- os.Hostname is used in the meepctl tool and does not work as expected in VM environment.
Reverted to using viper.Getstring.

- fixed meepctl ignoring build failures when dockerizing 

- moved kibana specific code in configKibana command


roymx:
The following should be addressed as part of this PR:
- move code from utils/kibana.go into cmd/configkibana.go
- meepctl ignores docker build failures

go-apps/meepctl/cmd/dockerize.go
```diff
@@ -127,47 +127,51 @@ func dockerize(registry string, targetName string, cobraCmd *cobra.Command) {
...
-	if registry != "" {
-		tag := registry + "/" + targetName
-		cmd := exec.Command("docker", "build", "--no-cache", "--rm", "--label", "MeepVersion="+checksum[0], "-t", tag, path)
-		_, _ = utils.ExecuteCmd(cmd, cobraCmd)
-		cmd = exec.Command("docker", "push", tag)
-		_, err := utils.ExecuteCmd(cmd, cobraCmd)
-		if err != nil {
-			fmt.Println("Failed to push", tag, " Error:", err)
-			return
+	if err == nil {
```
roymx: Follow golang practice of checking for error first makes code more readable and minimize levels.
e.g.
**do something**
if err != nil {
**manage error & return**
}
** no error -- continue processing**

go-apps/meepctl/cmd/dockerize.go
```diff
@@ -127,47 +127,51 @@ func dockerize(registry string, targetName string, cobraCmd *cobra.Command) {
...
+				return
+			}
+		} else {
+			cmd := exec.Command("docker", "build", "--no-cache", "--rm", "--label", "MeepVersion="+checksum[0], "-t", targetName, path)
+			_, _ = utils.ExecuteCmd(cmd, cobraCmd)
 		}
 	} else {
-		cmd := exec.Command("docker", "build", "--no-cache", "--rm", "--label", "MeepVersion="+checksum[0], "-t", targetName, path)
-		_, _ = utils.ExecuteCmd(cmd, cobraCmd)
+		fmt.Println("dockerizing could not be initiated: Error with the build for", targetName)
```
roymx: fmt.Println("Error dockerizing ", targetName)

go-apps/meepctl/cmd/dockerize.go
pastorsx: I changed the order to check != nil first but let the else in there to allow the code to then continue to get to a section where a clean up of some intermediate steps taken while building/dockerizing is occuring

go-apps/meepctl/cmd/dockerize.go
pastorsx: updated the error code message to match the suggestion.

PR #35, opened 2019-10-29
NA-660 (Event log), NA-701(Keep metrics polling when not in exec tab)

renaudmx:


renaudmx:
Cypress tests not run yet

PR #49, opened 2019-11-25
Adding influxdb helm chart

pastorsx:
New dependency (deployed when issues meepctl deploy dep) to enable support of influxdb.

Helm chart for InfluxDB added
   - supports persistence
   - meepctl updated (deploy, delete)
   - exposed through a nodeport

Basic manipulations of the db was tested.. adding a db, writing a value in the db, querying the value in the db (using curl). Deletion of the pod and recreation done to confirm persistence of data stored.


charts/influxdb/templates/service-nodeport.yaml
```diff
@@ -0,0 +1,24 @@
...
+  type: {{ .Values.serviceNodePort.type }}
+  ports:
+  {{- if .Values.config.http.enabled }}
+  - name: api
+    port: {{ .Values.config.http.bind_address }}
+    targetPort: {{ .Values.config.http.bind_address }}
+    nodePort: {{ .Values.serviceNodePort.nodeport }}
+  {{- end }}
+  selector:
+    app: {{ template "influxdb.fullname" . }}
```
dilallkx: Add Nodeport support to the influxDB service.yaml file instead of here. See other charts for examples.

charts/influxdb/values.yaml
```diff
@@ -0,0 +1,311 @@
...
+  # externalIPs: []
+  ## Specify LoadBalancer IP (only allow on some cloud provider)
+  # loadBalancerIP: ""
+  ## Allow source IPs to access on service (if empty, any access allow)
+  # loadBalancerSourceRanges: []
+
+serviceNodePort:
+  name: influxdb-nodeport
+  type: NodePort
+  nodeport: 31333
```
dilallkx: Integrate this in service instead. Service can either be ClusterIP or NodePort, not both. Make sure the templates handle this correctly.

dilallkx: Also, I recommend using port number 30086 for http service to match the redis & couchdb pattern of using the last 3 port digits.

go-apps/meepctl/cmd/delete.go
```diff
@@ -146,11 +146,13 @@ func deleteDep(cobraCmd *cobra.Command) {
 	go k8sDeletePvc("data-meep-elasticsearch-data-0", cobraCmd, messages)
 	go k8sDeletePvc("data-meep-elasticsearch-master-0", cobraCmd, messages)
 	go k8sDeletePvc("data-meep-elasticsearch-master-1", cobraCmd, messages)
+        go k8sDeletePvc("meep-influxdb-data-meep-influxdb-0", cobraCmd, messages)
```
dilallkx: You should not need this. It seems you are missing a PVC in your chart templates.

charts/influxdb/templates/service-nodeport.yaml
pastorsx: nodeport was added to service and file specific to nodeport was removed

charts/influxdb/values.yaml
pastorsx: using suggested ports and nodeport specific yaml file was removed

go-apps/meepctl/cmd/delete.go
pastorsx: removed.. now handled as part of a new pvc yaml file

PR #51, opened 2019-11-29
Metric Store with InfluxDB

dilallkx:
**CHANGES:**
- Created meep-metric-store package
  - Uses InfluxDB to store measurements/metrics
  - Manages DB connection and allows user to set & flush a specific store in the DB
  - Provides general functions to store and retrieve metrics in DB
  - Provides metric-specific functions to set & get network and event metrics
- Updated ctrl-engine to log events in metric store
- Updated tc-sidecar to log latency & traffic measurements in metric store
- Added MEEP_SCENARIO_NAME to tc-sidecar environment variables
- Added affinity to influxdb chart to schedule on master node

**TESTING:**
- Added & ran unit tests for new meep-metric-store package
- Cypress tests passed


go-packages/meep-metric-store/metric-store.go
```diff
@@ -0,0 +1,243 @@
...
+		return err
+	}
+
+	// logTimeLapse("SetMetric duration: ")
+
+	return nil
+}
+
+// GetMetric - Generic metric getter
+func (ms *MetricStore) GetMetric(metric string, tags map[string]string, fields []string, duration string, count int) (values []map[string]interface{}, err error) {
```
pastorsx: Isn't this a limiting factor to assume a duration always based on Now()-XX.. preventing from doing searches as part of a group of historical data.. ?

go-packages/meep-metric-store/metric-store_test.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		t.Errorf("Unable to set Store")
+	}
+	fmt.Println("Set store2")
+	err = ms.SetStore(metricStore2Name)
+	if err != nil {
+		t.Errorf("Unable to set Store2")
+	}
+
+	// t.Errorf("DONE")
```
pastorsx: remove these

go-packages/meep-metric-store/network.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	// Take first & only values
+	values := valuesArray[0]
+	tput = JsonNumToFloat64(values["tput"].(json.Number))
+	loss = JsonNumToFloat64(values["loss"].(json.Number))
+	return
+}
+
+// GetLatencyMetrics
```
pastorsx: not matching function name

go-packages/meep-metric-store/events.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Take first & only values
+	values := valuesArray[0]
+	if val, ok := values["event"].(string); ok {
+		event = val
+	}
+	return event, nil
+}
+
+// GetLatencyMetrics
```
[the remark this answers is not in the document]
dilallkx: Fixed.

go-packages/meep-metric-store/metric-store.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, err
+	}
+
+	log.Info("Successfully connected to Influx DB")
+	return ms, nil
+}
+
+func (ms *MetricStore) connectDB(addr string) error {
+	if addr == "" {
+		ms.addr = "http://meep-influxdb:8086"
```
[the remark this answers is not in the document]
dilallkx: Done.

go-packages/meep-metric-store/metric-store.go
```diff
@@ -0,0 +1,243 @@
...
+	log.Debug("InfluxDB Connector connecting to ", ms.addr)
+
+	client, err := influxclient.NewHTTPClient(influxclient.HTTPConfig{Addr: ms.addr, InsecureSkipVerify: true})
+	if err != nil {
+		log.Error("InfluxDB Connector unable to connect ", ms.addr)
+		return err
+	}
+	defer client.Close()
+
+	_, version, err := client.Ping(1000 * time.Millisecond)
```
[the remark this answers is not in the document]
dilallkx: The ping is already in a function with a few retries (connectDB).

go-packages/meep-metric-store/metric-store.go
dilallkx: Yes, this was done to simplify the interface for now. The API should be implemented the same way as well for now, at least until we have a need to time ranges. Another reason for this simplification is to avoid timezone differences & out of sync times.

go-packages/meep-metric-store/network.go
dilallkx: Fixed.

go-packages/meep-metric-store/metric-store_test.go
dilallkx: Done.

go-apps/meep-tc-sidecar/destination.go
```diff
@@ -130,182 +131,157 @@ func (u *destination) compute(rc *redis.Connector) (st stat) {
...
-	var throughputStats = make(map[string]interface{})
-	throughputStats[u.remoteName] = throughputVal
+	// Store latest values for next calculation
+	u.prevRx.time = curTime
+	u.prevRx.rxBytes = curRxBytes
 
-	//store statistics but only if the entry exists
-	key := moduleMetrics + ":" + PodName + ":" + u.remoteName
+	// Store throughput metric if entry exists
+	var tputStats = make(map[string]interface{})
```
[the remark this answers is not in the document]
dilallkx: We want to set a single throughput value here and the interface requires a map. This is why the map was created with a single entry. We should not push multiple throughput values here.

PR #58, opened 2020-01-10
frontend update (clone feature, vis fixes, editable attributes)

pastorsx:
Front end update:

1) Scenario element specificity for display:
Scenario element in frontend does not show the type and parent name anymore in editable mode.

2) Unique suggested name based on element type
When creating a new element, a suggested name appears based on the type of the element to create. If the type is modified, the suggested name also is modified accordingly. If the name already exists, a numerical number is added after the suggested name. New elements now have a unique uuid() upon creation. Older scenario not using the new scheme are backward compatible (assuming the id/name is unique, as it used to be the case but using the same as the name)

3) Editable element name
Names of elements already created are now editable. 

4) Clone functionality
CLONE button added in the edit panel. When pressed, clears the parent and suggest a name based on the element cloned. Upon pushing the APPLY button, it clones the tree underneath the selected element.

5) Edit panel button reshaping
Top buttons for the edit panel (NEW, DELETE, CLONE) have the same length and moved under the Configuration Element title

6) Vis window redraw
- Vis window redraws when the scenario is updated, but zoom level is kept as it was before the scenario changed.
- Fixed elements picture that were lost when zoom level got to a certain point in configure tab.
- Fixed continuous reinitialization of devMode (in Settings) causing attributes to be lost upon clicking on different parameters (Physics, manipulation, interaction, etc...)

Testing: multiple combinations of set, and execute scenarios.
Cypress run: updated and passedzoom


js-apps/meep-frontend/src/js/containers/cfg/cfg-page-container.js
```diff
@@ -141,6 +148,141 @@ class CfgPageContainer extends Component {
...
+    setElemFieldVal(newElement, FIELD_PARENT, newParentName);
+
+    // add new element to scenario
+    // new id and label will be created as part of the addNewElementToScenario called by newScenarioElem
+    this.props.newScenarioElem(newElement, false);
+    return name;
+  }
+
+  // CLONE
+  onApplyCloneElement(element) {
```
dilallkx: All scenario manipulations should be in the scenario-utils.js file. Please move this function and make sure the scenario state is updated in the correct order; i.e. update the scenario and set the updated state when completed successfully. This scenario state update should then trigger a refresh of the visualization.

js-apps/meep-frontend/src/js/containers/meep-container.js
```diff
@@ -323,8 +323,12 @@ class MeepContainer extends Component {
     }, 2000);
   }
 
-  // Change & process scenario
   changeScenario(pageType, scenario) {
+    changeScenario(pageType, scenario, false);
+  }
+
+  // Change & process scenario
+  changeScenario(pageType, scenario, reInitVisView) {
```
dilallkx: More recent linter does not accept functions with the same name. Suggest renaming one of the functions to avoid possible issues.

js-apps/meep-frontend/src/js/util/scenario-utils.js
```diff
@@ -236,25 +237,91 @@ export function parseScenario(scenario) {
   return { table: table, visData: visData };
 }
 
+function findIdInScenario(scenario, uniqueId) {
```
dilallkx: Is there a better way of determining if the ID is in the scenario? E.g. by keeping a map of IDs.

js-apps/meep-frontend/src/js/util/scenario-utils.js
```diff
@@ -417,12 +485,32 @@ export function updateElementInScenario(scenario, element) {
         element,
         FIELD_INT_ZONE_PKT_LOSS
       );
+
+      //if domain name changed, other elements created based on that name must also be updated (default ones)
+      for (var i2 in domain.zones) {
```
dilallkx: This seems like an update with potential for making the scenario invalid. Are there many cypress or unit tests to augment the coverage here?

js-apps/meep-frontend/src/js/util/scenario-utils.js
pastorsx: Only invoked when a new element scenario is created which should happen once and then not anymore. Although the looping through the scenario might seem extensive, checking the scenario for duplicated names is a security of its uniqueness. Keeping a global map of what ids are in the scenario could be useful, but not believed to be essential.

js-apps/meep-frontend/src/js/containers/meep-container.js
pastorsx: updated accordingly

js-apps/meep-frontend/src/js/containers/cfg/cfg-page-container.js
pastorsx: moved to proper location

js-apps/meep-frontend/src/js/containers/cfg/cfg-network-element-container.js
```diff
@@ -1061,6 +1120,9 @@ export class CfgNetworkElementContainer extends Component {
                   onDeleteElement={() => {
                     this.props.onDeleteElement(element);
                   }}
+                  onCloneElement={() => {
+                    this.onCloneElement(createUniqueName(this.props.tableData, getElemFieldVal(element, FIELD_NAME) + '-copy'));
```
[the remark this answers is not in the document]
pastorsx: idea was to start cloning with a name already given... different function to create a unique name could then be used. I believe it is more generic that way from the point of view of onCLoneElement function. No change done.

js-apps/meep-frontend/src/js/util/scenario-utils.js
pastorsx: cypress cloning test imply clicking in the screen to select an object... was not looked at so no cypress added for that or unit tests

PR #60, opened 2020-01-30
logs optimizations

pastorsx:
Few log optimizations
1) logs are by default set in text mode, not in JSON anymore
2) logs extended to show the time of the system in us, file and line where the debug log was invoked from
3) every HTTP request is now logged at debug level
4) every event send or received from a channel is logged at Trace level from the redis package
5) every events send or received from a channel that was considered useful is logged at debug level prior to calling redis publish()
6) sidecar logs used for latencies removed (covered by influxDB)
7) call to redis to set an entry removed from a loop (ProcessRxTx) to outside the look for one write per cycle, rather than N write (N being the number of pods-1)
8) new Trace mode for extended debug
9) debugs added to show how much TC was busy on sidecars with performance metrics (duration of execution)

### DIFF
--- a/go-packages/meep-logger/logger.go
+++ b/go-packages/meep-logger/logger.go
@@ -17,11 +17,10 @@
 package logger
 
 import (
-	"net/http"
-	"time"
-	"runtime"
-	"path"
 	"fmt"
+	"path"
+	"runtime"
+
 	logrus "github.com/sirupsen/logrus"
 )
 
@@ -30,15 +29,19 @@ var componentName string
 type Fields map[string]interface{}
 
 func MeepTextLogInit(name string) {
+	Formatter := new(logrus.TextFormatter)
+	Formatter.TimestampFormat = "2006-01-02T15:04:05.999999999Z07:00"
+	Formatter.FullTimestamp = true
+	logrus.SetFormatter(Formatter)
 	//logrus.SetLevel(logrus.TraceLevel)
-        logrus.SetLevel(logrus.DebugLevel)
+	logrus.SetLevel(logrus.DebugLevel)
 	componentName = name
 }
 
 func MeepJSONLogInit(name string) {
 	logrus.SetFormatter(&logrus.JSONFormatter{})
 	//logrus.SetLevel(logrus.TraceLevel)
-        logrus.SetLevel(logrus.DebugLevel)
+	logrus.SetLevel(logrus.DebugLevel)
 	componentName = name
 }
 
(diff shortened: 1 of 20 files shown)


go-packages/meep-logger/logger.go
```diff
@@ -17,59 +17,117 @@
...
 
+// getLogCaller
+// setReportCaller cannot ignore caller levels, so cannot work in a wrapper. Fix not merged in github, so doing it manually
+func getLogCaller() string {
+	_, file, line, _ := runtime.Caller(2)
+
+        location := fmt.Sprintf("%v:%v", path.Base(file), line)
+	return location
+}
+
```
dilallkx: Why not use the built-in logrus method to enable file logging?
https://godoc.org/github.com/sirupsen/logrus#SetReportCaller

Also, you can format the output as you wish using the CallerPrettyfier:
https://github.com/sirupsen/logrus/issues/63

go-packages/meep-logger/logger.go
```diff
@@ -17,59 +17,117 @@
...
 func Info(args ...interface{}) {
-	log.WithFields(log.Fields{
+	logrus.WithFields(logrus.Fields{
 		"meep.component": componentName,
+		"meep.time": time.Now().String(),
```
dilallkx: Is this timestamp in addition to the default timestamp from logrus? If so, we should not have 2 timestamps.

Also, can't you use the logrus TimestampFormat & formatters to set a more precise full timestamp?

go-packages/meep-logger/logger.go
```diff
@@ -17,59 +17,117 @@
...
+		start := time.Now()
+
+		inner.ServeHTTP(w, r)
+
+		httpLog(
+			r.Method,
+			" ", r.RequestURI,
+			" apiFct: ", name,
+			" execTime: ", time.Since(start))
+	})
```
dilallkx: Why is a logging package serving HTTP? This should probably not be in this package.

go-packages/meep-logger/logger.go
```diff
@@ -17,59 +17,117 @@
...
 
 var componentName string
 
 type Fields map[string]interface{}
 
 func MeepTextLogInit(name string) {
-	log.SetFormatter(&log.TextFormatter{})
-	log.SetLevel(log.DebugLevel)
+	//logrus.SetLevel(logrus.TraceLevel)
+        logrus.SetLevel(logrus.DebugLevel)
```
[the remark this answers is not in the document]
pastorsx: fixed

go-packages/meep-logger/logger.go
pastorsx: this method is good but only comes back one level. As per comment, the reason why it was done so is because since we use a wrapper, we need to go up 2 calls in the stack.. otherwise the log will not be of the caller of our wrapper but from our wrapper itself.. There is a need to call caller(2) in this case as describe in: https://github.com/sirupsen/logrus/issues/972

go-packages/meep-logger/logger.go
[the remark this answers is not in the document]
pastorsx: fixed

go-packages/meep-logger/logger.go
pastorsx: The setReportCaller only goes up one level in the stack. As per the comment at the beginning of the function, since we are in a wrapper, we need to go up 2 level in the stack.. which is why the setReportCaller couldn't be used. If used, it would show the line and file of our wrapper as per https://github.com/sirupsen/logrus/issues/972

go-packages/meep-logger/logger.go
pastorsx: the reason for both timestamp was actually the fact that we wanted more precision, I will check

go-packages/meep-logger/logger.go
pastorsx: changed to have logger created from swagger to call meepLogger

go-packages/meep-logger/logger.go
pastorsx: using the formatter.. but its either seconds or ns... i then set it to ns precision, only one time required

dilallkx: Formatter seems to do the job. Also it looks like you can set microsecond granularity using six 0 or 9 after the point.

pastorsx: true... it is possible to use finer granularity... was changed to ms, although we could have had us or ns... can always be changed in the future if more granularity is required, but ms seems appropriate.

PR #61, opened 2020-02-03
meep couch package

roymx:
A wrapper package around CouchDB kivik driver that allows to have multiple document stores to the same couch instance.


go-apps/meep-ctrl-engine/go.sum
```diff
@@ -1,3 +1,4 @@
+github.com/InterDigitalInc/AdvantEDGE v1.3.0 h1:c7rr1cwOWXsTYyYLQ0+DFUAWqxarqMFMrPGoi1AU84o=
```
dilallkx: This dependency should not be here.

dilallkx: Removed in latest commit.

PR #62, opened 2020-02-04
Grafana Integration + Elastic Stack Deprecation

dilallkx:
**CHANGES:**
- Added meep-grafana helm chart
- meepctl:
  - Added meep-grafana to dependencies
  - Removed curator, filebeat, metricbeat, elasticsearch, kibana & ESK persistent volumes
  - Bumped repo config & meepctl versions from 1.3.0 --> 1.3.1
- Metric Store:
  - Added description field in Event metrics
  - Updated unit tests
  - Added method to copy Influx database measurements
  - Fixed value validation errors in network measurements
- Remove metrics engine V1 server
- Fixed concurrency issue in meep-model package
  - Added mutex on package API
- Updated IDCC icon
- Frontend:
  - Adjusted visualization pane heights
  - Removed unused d3 throughput & latency graphs
  - Added metrics view Grafana dashboard with dynamic src & dest app configuration
  - Removed unused config points from dashboard configuration
  - Added buttons in Exec tab to open/close dashboard configuration
  - Moved 'show apps' checkbox into dasshboard config menu
  - Aligned buttons on monitoring tab
  - Replaced link to Kibana by a link to Grafana UI

**TESTING:**
- Cypress tests updated
- Cypress & Unit tests pass


charts/couchdb/templates/statefulset.yaml
```diff
@@ -63,20 +63,6 @@ spec:
 {{- end }}
             - name: ERL_FLAGS
               value: "{{ range $k, $v := .Values.erlangFlags }} -{{ $k }} {{ $v }} {{ end }}"
```
roymx: Thought -- Still wondering if we should keep-it and make timeout longer... (how much longer?)

charts/grafana/ci/with-dashboard-json-values.yaml
```diff
@@ -0,0 +1,53 @@
+dashboards:
+  my-provider:
+    my-awesome-dashboard:
```
roymx: Do we really need my-awesome-dashboard?

charts/grafana/ci/with-dashboard-values.yaml
```diff
@@ -0,0 +1,19 @@
...
+  my-provider:
+    my-awesome-dashboard:
+      gnetId: 10000
+      revision: 1
+      datasource: Prometheus
+dashboardProviders:
+  dashboardproviders.yaml:
+    apiVersion: 1
+    providers:
+    - name: 'my-provider'
```
roymx: If this section is used, should we rename my-provider to advantedge-provider?

go-apps/meep-ctrl-engine/server/ctrl-engine.go
```diff
@@ -817,9 +830,9 @@ func sendEventPoasInRange(event ceModel.Event) (error, int) {
 		}
 	} else {
 		err := errors.New("Failed to find UE")
-		return err, http.StatusNotFound
+		return err, http.StatusNotFound, ""
 	}
-	return nil, -1
+	return nil, -1, ""
```
roymx: This one should have a description too.

charts/couchdb/templates/statefulset.yaml
dilallkx: Let's omit them for now to avoid issues with slower systems. When we fix the pod bring-up dependencies we can re-introduce the probes.

go-apps/meep-ctrl-engine/server/ctrl-engine.go
dilallkx: Done.

charts/grafana/ci/with-dashboard-json-values.yaml
dilallkx: These are from the original helm chart which I did not modify.

charts/grafana/ci/with-dashboard-values.yaml
dilallkx: These are from the original helm chart which I did not modify.

PR #64, opened 2020-02-14
Replay feature : api and backend

pastorsx:
CHANGES:
Ctrl-Engine:
Support for Replay Manager that plays Replay files

REST API updates (Replay endpoint added) : 
- get all replay files (shown by name)
- get a replay file by name
- delete all replay files
- delete a replay file by name
- create a replay file based on body of a request
- create a replay file based on history of events that happened during a scenario execution
- play a replay file once
- play a replay file in an infinite loop
- stop a replay file

CouchDB package::
Modified to return docs by name (ID)


dilallkx:
This PR will be included in PR #65

PR #68, opened 2020-02-28
Replay feature frontend

pastorsx:
New addition:
- Event button instead of Create Event button in exec toolbar
- Replay toolbar (manual, auto, create replayfile)
- create event pane invoked by replay toolbar
- replay pane


dilallkx:
This PR is included in PR #70.

PR #74, opened 2020-04-01
Ingress Controller

dilallkx:
**CHANGES:**
- Ingress:
  - Added nginx-ingress helm chart (k8s-supported ingress controller)
  - Configured ingress controller to use host network and listen on ports 80 & 443 for HTTP/HTTPS
  - Added Ingress resources to core & dependency charts templates
  - Removed unused NodePorts from core & dependency microservices
  - Changed Nodeport references in code to use ingress paths instead
  - Remaining Nodeports:
     - 30001: Docker registry
     - 30379: Redis DB, required for Virt Engine running outside k8s cluster
     - 30984: Couch DB, required to expose CouchDB Frontend
  - Services now accessible using following paths:
     - /: AdvantEDGE Frontend
     - /api: Swagger UI Frontend
     - /ctrl-engine: Ctrl Engine REST API
     - /grafana: Grafana Frontend 
     - /influxdb: Influx DB http API 
     - /location: ETSI mec location service REST API
     - /metrics: Metrics Engine REST API
     - /mgm: Mobility Group Manager REST API (App state transfer REST API)
- REST APIs:
  - Removed nodeport information from specs
  - Changed Location service basepath to '/location/v1'
  - Set Core services basepath to match ingress paths listed above
- Location & Metrics Services:
  - Added the following environment variables to specify resource URL when service is created:
     - MEEP_LOC_SERV_ROOT_URL
     - MEEP_METRICS_ROOT_URL
- meepctl:
  - Added meep-ingress to list of dependency microservices
  - Added self-signed TLS certificates for ingress controller
  - Added ports configuration entry to allow user to configure Nodeports instead of default 80/443 ports
- Tests:
  - Adapted tests to access AdvantEDGE frontend on port 80 by default

**TESTING:**
- Cypress tests pass using http & https


.meepctl-repocfg.yaml
```diff
@@ -180,16 +180,17 @@ repo:
...
-      chart: charts/docker-registry
     couchdb:
       chart: charts/couchdb
+    docker-registry:
+      chart: charts/docker-registry
+    grafana:
+      chart: charts/grafana
     influxdb:
       chart: charts/influxdb
+    kube-state-metrics:
```
roymx, line 191: Why did we add kube-state-metrics again?
An ingres controller dependency?

charts/meep-webhook/templates/service.yaml
```diff
@@ -16,7 +16,8 @@ spec:
...
     meepOrigin: {{ .Values.meepOrigin }}
   ports:
-    - port: {{ .Values.deployment.port }}
-      targetPort: {{ .Values.service.targetPort }}
+    - port: {{ .Values.service.port }}
+      targetPort: {{ .Values.deployment.port }}
+      {{- if .Values.service.nodePort }}
       nodePort: {{ .Values.service.nodePort }}
-
+      {{- end }}
```
roymx, line 23: I think you need a newline at the end of a yaml file

charts/meep-webhook/values.yaml
```diff
@@ -15,9 +15,10 @@ deployment:
   dependency: meep-redis-master
 
 service:
-  type: NodePort
-  targetPort: 443
-  nodePort: 30006
+  type: ClusterIP
+  # type: NodePort
+  port: 443
+  # nodePort: 30006
```
roymx, line 21: I note you kept the old ports in comment.
I think these can be safely removed now.

.meepctl-repocfg.yaml
dilallkx: I did not add kube-state-metrics here. I just moved it and removed the k8s tag.
Although kube-state-metrics is not currently used, it is available for Grafana dashboards to get metrics from the Kubernetes API.

charts/meep-webhook/templates/service.yaml
dilallkx: fixed.

charts/meep-webhook/values.yaml
dilallkx: Removed.

PR #88, opened 2020-05-12
Location service UT testing

pastorsx:
Location Service UT testing (based on rnis model for UT testing that was previously added)

### DIFF
--- a/go-apps/meep-loc-serv/go.sum
+++ b/go-apps/meep-loc-serv/go.sum
@@ -1,4 +1,5 @@
 cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+github.com/InterDigitalInc/AdvantEDGE v1.4.0 h1:MnzK3dl4hUvfhZ4wYarVmJzE5IMO2Q63dyDkcQtEv1Q=
 github.com/KromDaniel/jonson v0.0.0-20180630143114-d2f9c3c389db/go.mod h1:RU+6d0CNIRSp6yo1mXLIIrnFa/3LHhvcDVLVJyovptM=
 github.com/KromDaniel/rejonson v0.0.0-20180822072824-00b5bcf2b351 h1:1u1XrfCBnY+GijnyU6O1k4odp5TnqZQTsp5v7+n/E4Y=
 github.com/KromDaniel/rejonson v0.0.0-20180822072824-00b5bcf2b351/go.mod h1:HxwfbuElTuGf+/uKZfjJrCnv0BmmpkPJDI7gBwj1KkM=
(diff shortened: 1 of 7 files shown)


go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -1233,6 +1239,13 @@ func cleanUp() {
 	userSubscriptionMap = map[int]string{}
 
 	zoneStatusSubscriptionMap = map[int]*ZoneStatusCheck{}
+
+	updateStoreName("")
+}
+
+func updateStoreName(storeName string) {
+	currentStoreName = storeName
+	_ = httpLog.ReInit(logModuleLocServ, sandboxName, storeName, redisAddr, influxAddr)
```
dilallkx, line 1248: This will be called on every active scenario model update, which could happen very often. You should check if the store name changed before calling the httpLog reinit function.

go-apps/meep-loc-serv/sbi/loc-serv-sbi.go
```diff
@@ -159,17 +161,17 @@ func processActiveScenarioUpdate() {
 		uePerNetLocMap[netLoc]++
 	}
 
-	// Update POA info
-	poaNameList := sbi.activeModel.GetNodeNames("POA")
+	// Update POA-CELL info
+	poaNameList := sbi.activeModel.GetNodeNames("POA-CELL")
```
dilallkx, line 165: RNIS requires Cellular POAs. Is the same true for the Location Service? If not then we should support both generic & cellular POAs here.

dilallkx: Does this change break cypress tests?

pastorsx: no change to cypress... this is only on which POA type location service check before increasing increasing the number of POAs in status report... I on purpose only put the cellular ones... This wasnot updated after cellular type was added. Location service is only meant for cellular POAs

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: true .. updated for both loc-serv and rnis

go-apps/meep-loc-serv/go.sum
```diff
@@ -1,4 +1,5 @@
 cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+github.com/InterDigitalInc/AdvantEDGE v1.4.0 h1:MnzK3dl4hUvfhZ4wYarVmJzE5IMO2Q63dyDkcQtEv1Q=
```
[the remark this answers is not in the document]
pastorsx: removed

PR #97, opened 2020-06-25
Network Characteristics v1.5 Updates

pastorsx:
- Network Characteristics object used throughout the data model
- Throughput split into two distinct field: Uplink and Downlink throughputs
- Latency distribution added to the network characteristics (editable only under the Scenario.Deployment object)

Tests:
cypress ran and passed
UT : passes


js-apps/meep-frontend/src/js/components/helper-components/nc-group.js
```diff
@@ -236,43 +256,52 @@ const NCGroup = ({ prefix, onUpdate, element }) => {
   // Retrieve field names
   var latencyFieldName = null;
   var latencyVarFieldName = null;
-  var throughputFieldName = null;
+  var latencyDistFieldName = null;
+  var throughputDlFieldName = null;
+  var throughputUlFieldName = null;
   var packetLossFieldName = null;
   switch (prefix) {
```
roymx, line 263: Question - since we moved to a common structure for net.chars
What is the benefit of keeping this switch case of specific net.char types?

go-packages/meep-net-char-mgr/algo-segment.go
```diff
@@ -208,14 +208,22 @@ func (algo *SegmentAlgorithm) ProcessScenario(model *mod.Model) error {
...
-			element.ConfiguredNetChar.Throughput = DEFAULT_THROUGHPUT_LINK
+		element.ConfiguredNetChar.Latency = float64(proc.NetChar.Latency)
+		element.ConfiguredNetChar.Jitter = float64(proc.NetChar.LatencyVariation)
+		element.ConfiguredNetChar.Distribution = deployment.NetChar.LatencyDistribution //set global value
+		element.ConfiguredNetChar.ThroughputDl = float64(proc.NetChar.ThroughputUl)
+		element.ConfiguredNetChar.ThroughputUl = float64(proc.NetChar.ThroughputUl)
+		element.ConfiguredNetChar.PacketLoss = float64(proc.NetChar.PacketLoss)
+
+		if element.ConfiguredNetChar.ThroughputUl == 0 {
+			element.ConfiguredNetChar.ThroughputUl = DEFAULT_THROUGHPUT_LINK
```
roymx, line 222: I'm not sure why Throughput=0 case is not supported and reverts to default.
In an upcoming feature, when we get to simulate a disconnected UE, throughput = 0 will need to be supported

go-packages/meep-net-char-mgr/algo-segment.go
```diff
@@ -400,13 +416,22 @@ func (algo *SegmentAlgorithm) populateFlow(flowName string, srcElement *SegAlgoN
...
 	if maxBw == 0 {
-		if srcElement.ConfiguredNetChar.Throughput < destElement.ConfiguredNetChar.Throughput {
-			maxBw = srcElement.ConfiguredNetChar.Throughput
+		if srcElement.ConfiguredNetChar.ThroughputUl < destElement.ConfiguredNetChar.ThroughputDl {
+			maxBw = srcElement.ConfiguredNetChar.ThroughputUl
 		} else {
-			maxBw = destElement.ConfiguredNetChar.Throughput
+			maxBw = destElement.ConfiguredNetChar.ThroughputDl
 		}
+/*                if srcElement.ConfiguredNetChar.ThroughputDl < destElement.ConfiguredNetChar.ThroughputUl {
```
roymx, line 424: Remove dead code

go-packages/meep-net-char-mgr/algo-segment.go
```diff
@@ -997,64 +1023,41 @@ func updateMaxFairShareBwPerFlow(segment *SegAlgoSegment) {
...
+		nc = deployment.NetChar
 	} else {
 		log.Error("Error casting element: " + elemName)
 	}
 
 	// For compatiblity reasons, set to default value if 0
-	if maxThroughput == 0 {
-		maxThroughput = DEFAULT_THROUGHPUT_LINK
+	if nc.ThroughputUl == 0 {
+		nc.ThroughputUl = DEFAULT_THROUGHPUT_LINK
```
roymx, line 1056: Same comment as before - throughput = 0 will need to be supported.
It is unclear why 0 defaults to 1Mbps

go-packages/meep-net-char-mgr/net-char-mgr.go
```diff
@@ -59,6 +59,17 @@ type NetChar struct {
 	Jitter     float64
 	PacketLoss float64
 	Throughput float64
+	Distribution string
+}
+
+// NetChar
+type ElemNetChar struct {
+        Latency    float64
+        Jitter     float64
```
roymx, line 68: This is inconsistent.
LatencyDistribution is used everywhere in the model, frontend, etc. and here we switch to Jitter

go-packages/meep-net-char-mgr/algo-segment.go
pastorsx: At some point, we said a throughput of 0 is no bandwidth limitation, and since the default is at 1GBps, that's why it was set to that value. we said at this point that we could use 100% packet loss too, so that is why it was like that. Either way, it's only on ingress side that the traffic will be cut, either because of bw 0 or packet loss of 100. I don't think one is better than the other, TBD.

js-apps/meep-frontend/src/js/components/helper-components/nc-group.js
pastorsx: Actually, this is a valid comment, the model used to be flattened, but it is not required anymore

go-packages/meep-net-char-mgr/algo-segment.go
pastorsx: same reply as above

go-packages/meep-net-char-mgr/algo-segment.go
pastorsx: removed

go-packages/meep-net-char-mgr/net-char-mgr.go
pastorsx: yes... jitter is used mainly in the net-char-package instead of latencyVariation. I agree this might bring confusion. Should this one be moved to latency variation or model should be changed to jitter first?

PR #99, opened 2020-06-26
Net Char Update Event fixes

dilallkx:
**CHANGES:**
- Net Char Event REST API:
  - Changed cloud type from 'DISTANT CLOUD' to 'DC' to match scenario model type (this was the only one out of sync)
- Model package:
  - Updated model types to match REST API/Scenario model types
  - Updated UT
- Frontend:
  - Adapted Event generation code to send the correct Scenario element types to backend (to be in line with REST API spec)

**TESTING:**
- Cypress & UT pass
- Verified manual event generation + auto-replay using frontend


js-apps/meep-frontend/src/js/components/dialogs/id-dialog.js
```diff
@@ -28,7 +28,7 @@ import {
 const IDDialog = props => {
   return (
     <Dialog
-      style={{zIndex: 1000}}
+      style={{zIndex: 10000}}
       open={props.open}
```
roymx, line 32: Not sure how zIndex 10000 relates to this bugfix.
Double-check if it's intended.

dilallkx: This is a minor frontend fix that allows the dialogs to appear above the map control buttons. It was intended but not mentioned in the PR.

PR #100, opened 2020-06-27
User specific configuration

roymx:
Support user configuration value files for user specific deployments
- Added configuration params in repocfg.yaml
- meep-virt-engine : added PV for user values files & use file if present
- meepctl : use user value file if present
- supported by core, sandbox & dep 

Other
- Blocked un-necessary nodeports (couchdb, open-map-tiles, redis)

### DIFF
--- a/.meepctl-repocfg.yaml
+++ b/.meepctl-repocfg.yaml
@@ -46,6 +46,8 @@ repo:
         bin: bin/meep-mon-engine
         # location of deployment chart
         chart: charts/meep-mon-engine
+        # user supplied value file located @ .meep/user/values
+        chart-user-values: meep-mon-engine.yaml
         # extra build flags
         build-flags:
           - -mod=vendor
@@ -94,6 +96,8 @@ repo:
         bin: bin/meep-platform-ctrl
         # location of deployment chart
         chart: charts/meep-platform-ctrl
+        # user supplied value file located @ .meep/user/values
+        chart-user-values: meep-platform-ctrl.yaml
         # enable meepctl build
         build: true
         # enable meepctl dockerize
@@ -119,6 +123,8 @@ repo:
         bin: bin/meep-virt-engine
         # location of deployment chart
         chart: charts/meep-virt-engine
+        # user supplied value file located @ .meep/user/values
+        chart-user-values: meep-virt-engine.yaml
         # enable meepctl build
         build: true
         # enable meepctl dockerize
@@ -158,6 +164,8 @@ repo:
         bin: bin/meep-webhook
         # location of deployment chart
         chart: charts/meep-webhook
+        # user supplied value file located @ .meep/user/values
+        chart-user-values: meep-webhook.yaml
         # extra build flags
         build-flags:
           - -mod=vendor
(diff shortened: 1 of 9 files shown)


go-apps/meepctl/cmd/deploy.go
```diff
@@ -193,6 +201,15 @@ func deployCore(cobraCmd *cobra.Command) {
...
 		}
+		if userValues {
+			// user provided overriding values
+			// Note: according to https://helm.sh/docs/chart_template_guide/values_files/
+			//       the order of precedence is: (lowest) default values.yaml
+			//                                            then user value file
+			//                                            then individual --set params (highest)
+			//       Therefore, the --set flags inserted by meepctl may interfere with user overrides
+			coreFlags = utils.HelmFlags(coreFlags, "-f", userValueFile)
+		}
```
dilallkx, line 212: The user values file is added in the function deployRunScriptsAndGetFlags(). Adding the flag here should not be necessary.

.meepctl-repocfg.yaml
```diff
@@ -46,6 +46,8 @@ repo:
         bin: bin/meep-mon-engine
         # location of deployment chart
         chart: charts/meep-mon-engine
+        # user supplied value file located @ .meep/user/values
+        chart-user-values: meep-mon-engine.yaml
```
dilallkx, line 50: This configuration point (i.e. chart-user-values) seems to be ignored in meepctl. The microservice name is used to identify the user-provided values.yaml file instead of the filename provided here. Is the intent to have this file name configurable here?

NOTE: The same question applies for sandbox microservices; the sandbox microservice names provided in the pod list environment variable are used to identify the user-provided values.yaml.

If they are not configurable we could simply document it here without adding configurability.

charts/couchdb/values.yaml
```diff
@@ -81,9 +81,10 @@ podManagementPolicy: Parallel
 ## to a second Service that governs how clients connect to the CouchDB cluster.
 service:
   enabled: true
-  type: NodePort
```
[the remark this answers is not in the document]
roymx: The idea here is to close the port to make the platform more secure by default.
A developer/user that needs to enable couch FE in their deployment can easily do so by using the current feature.
e.g. create a meep-couchdb.yaml in `.meep/user.values` that contains this:
```
service:
    type: NodePort
    nodePort: 30984
```

go-apps/meepctl/cmd/deploy.go
roymx: Right... moved it but forgot to clean this one.
Will update

.meepctl-repocfg.yaml
roymx: For this first version: not configurable.
Main reason is that meepctl & repo-config are not used in the virt-engine.
When/If we change virt-engine approach to leverage meepctl ... then we can really make these configurable
For now, I will clarify this in repocfg.

charts/couchdb/values.yaml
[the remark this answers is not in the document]
dilallkx: Makes perfect sense. Thanks!

PR #108, opened 2020-07-14
Chart Fixes

dilallkx:
**CHANGES:**
- Disabled InfluxDB ingress
- Fixed swagger UI ingress rules to append trailing slash on /api
- Removed unused grafana values.yaml override
- Disabled Swagger UI validator

**TESTING:**
- Cypress & UT pass

### DIFF
--- a/charts/influxdb/values.yaml
+++ b/charts/influxdb/values.yaml
@@ -111,7 +111,7 @@ resources: {}
 podAnnotations: {}
 
 ingress:
-  enabled: true
+  enabled: false
   tls: false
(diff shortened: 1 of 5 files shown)


dilallkx:
> What was the reason for deploy.go changes?

The Grafana values.yaml file had to be updated with the cluster IP address when it was deployed in direct DB access mode. Now that it is in proxy mode it no longer needs this modification. Also, setting this values.yaml file in meepctl prevented the user values override functionality from working for grafana.

PR #118, opened 2020-08-18
Configurable zone colors (Frontend)

dilallkx:
**CHANGES:**
- New zone color selection field in scenario configuration
  - Accepts HEX formatted input
  - Displays selected color in an icon button
  - Enables ChromePicker color selector on color icon button click
    - requires new package dependency on _react-color_

**TESTING:**
- Cypress & UT pass
- Manual testing of zone color configuration

### DIFF
--- a/js-apps/meep-frontend/src/js/util/elem-utils.js
+++ b/js-apps/meep-frontend/src/js/util/elem-utils.js
@@ -135,8 +135,18 @@ export const setElemFieldErr = (elem, field, err) => {
   }
 };
 
+export const resetElem = (elem) => {
+  if (elem) {
+    elem.editColor = false;
+  }
+};
+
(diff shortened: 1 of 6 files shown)


js-apps/meep-frontend/src/js/util/elem-utils.js
```diff
@@ -135,8 +135,18 @@ export const setElemFieldErr = (elem, field, err) => {
   }
 };
 
+export const resetElem = (elem) => {
```
pastorsx, line 138: Does the resetElem only have use for resting the color editor part? 
If so, shouldn't we rename it for its only purpose.  (If it only restes states but only one state exists, should be called resetElemStates)
If not, should we had other data that needs to be reseted

PR #119, opened 2020-08-19
Added DNS lookup with init container for pods in core platform

nikhildoifode:


nikhildoifode:
https://stackoverflow.com/questions/52109039/nslookup-cant-resolve-kubernetes-default
https://github.com/kubernetes/dns/issues/253

PR #122, opened 2020-08-26
Retention of Manual Event Creation fields on changing tabs

nikhildoifode:
Changes:
1. Added '@' as alias for src folder for having absolute path while importing
2. Added Mobility Event target and destination to state for retention
3. Disabled loading Mobility event as default
4. Added clearing all fields funtionality on clicking Close button
5. Disabled local state in Component and changed it to global state
6. Added value field to Tags where it was missing.

Testing:
Cypress & UT pass
Manual testing of Event Fields and changing the values


js-apps/meep-frontend/src/js/containers/exec/mobility-event-pane.js
```diff
@@ -24,19 +25,24 @@ import { EXEC_EVT_MOB_TARGET, EXEC_EVT_MOB_DEST } from '../../meep-constants';
...
+
 class MobilityEventPane extends Component {
   constructor(props) {
     super(props);
-    this.values = {
-      eventType: '',
-      eventTarget: '',
-      eventDestination: ''
-    };
+    // this.values = {
```
dilallkx, line 36: General comment: No need to keep old code commented out unless it has value. Suggest removing unnecessary comments.

js-apps/meep-frontend/webpack.config.js
```diff
@@ -23,6 +23,12 @@ module.exports = env => {
     node: {
       fs: 'empty'
     },
+    resolve: {
+      extensions: ['.js', '.json'],
+      alias: {
+        '@': path.resolve('src')
```
dilallkx, line 29: does this impact imports like '@rmwc'?

js-apps/meep-frontend/src/js/containers/exec/network-characteristics-event-pane.js
```diff
@@ -131,14 +131,21 @@ class NetworkCharacteristicsEventPane extends Component {
...
-      dialogOpen: false,
-      currentElementType: ''
+      dialogOpen: false
+      // currentElementType: ''
     };
   }
 
+  onNetworkCharacPaneClose(e) {
+    e.preventDefault();
+    setElemFieldVal(this.props.element, FIELD_TYPE, '');
```
dilallkx, line 141: We should not update the selected element state directly here. Update a copy & dispatch instead. Could we clear the selected element instead of just changing its type?

js-apps/meep-frontend/src/js/containers/exec/network-characteristics-event-pane.js
```diff
@@ -385,9 +398,10 @@ class NetworkCharacteristicsEventPane extends Component {
               onChange={event => {
                 var elem = this.firstElementMatchingType(event.target.value);
                 this.props.updateElement(elem);
-                this.setState({ currentElementType: event.target.value });
+                // this.setState({ currentElementType: event.target.value });
```
dilallkx, line 401: We should clear the selected element if not matches the selected type.

js-apps/meep-frontend/src/js/containers/exec/network-characteristics-event-pane.js
```diff
@@ -131,14 +131,21 @@ class NetworkCharacteristicsEventPane extends Component {
     super(props);
 
     this.state = {
-      dialogOpen: false,
-      currentElementType: ''
+      dialogOpen: false
+      // currentElementType: ''
```
dilallkx, line 135: You may need to add a state for the element type select menu. Other option would be to filter menu options to list only supported types.

js-apps/meep-frontend/webpack.config.js
nikhildoifode: It won't affect other imports.

js-apps/meep-frontend/src/js/containers/exec/mobility-event-pane.js
nikhildoifode: Removed the commented code

js-apps/meep-frontend/src/js/containers/exec/network-characteristics-event-pane.js
nikhildoifode: Updated to change the element in store

js-apps/meep-frontend/src/js/containers/exec/network-characteristics-event-pane.js
nikhildoifode: Added code to update the element type list based on scenario deployed

js-apps/meep-frontend/src/js/containers/exec/network-characteristics-event-pane.js
nikhildoifode: Selected element will automatically get updated now since we are updating the list based on scenario deployed.

dilallkx (approved):

PR #124, opened 2020-08-28
Session Keepalive

dilallkx:
**CHANGES:**
- New keepalive endpoint (GET /login) for FE session heartbeat
  - Session timestamp & cookie updated on login & on each keepalive request
- New Session watchdog to remove timed out sessions
  - Runs every minute and removes old sessions
  - Invokes handler on session removal to trigger sandbox destruction
- Session cookie & wathdog session timeout set to 20 minutes

**TESTING:**
- Cypress & UT pass
- Manually tested:
  - Session keepalive endpoint
  - Session & sandbox cleanup on timeout


go-apps/meep-platform-ctrl/api/swagger.yaml
```diff
@@ -25,6 +25,21 @@ produces:
 - "application/json"
 paths:
   /login:
```
roymx, line 27: Not a huge thing 
- would have expected an endpoint of its own for the watchdog (e.g. /watchdog) and the user to POSTing to that endpoint (e.g. user kicking the watchdog)
- whether the "kicking" process generates a new cookie or not is internal to the platform and should not necessarily be exposed through the API

js-packages/meep-platform-ctrl-client/test/api/UserAuthenticationApi.spec.js
```diff
@@ -93,6 +93,24 @@
           done();
         });
       });
+      describe('refreshUser', function() {
+        it('should call refreshUser successfully', function(done) {
+          // TODO: uncomment refreshUser call
```
roymx, line 98: Unclear why this is commented.

go-apps/meep-platform-ctrl/server/user_authentication.go
```diff
@@ -140,3 +140,18 @@ func uaLogoutUser(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
 	w.WriteHeader(http.StatusOK)
 }
+
+func uaRefreshUser(w http.ResponseWriter, r *http.Request) {
+	log.Info("----- KEEP-ALIVE -----")
```
roymx, line 145: We should remove this log... it will be noisy.

go-packages/meep-sessions/session-mgr.go
```diff
@@ -114,3 +125,56 @@ func (sm *SessionMgr) Authorizer(inner http.Handler) http.Handler {
...
+						continue
+					}
+
+					// Remove timed out sessions
+					currentTime := time.Now()
+					for _, session := range sessionList {
+						if currentTime.After(session.Timestamp.Add(SessionDuration * time.Second)) {
+							// Invoke watchdog timeout handler
+							if sm.wdHandler != nil {
+								sm.wdHandler(session)
```
roymx, line 156: - Not invoking sm.wdHandler here could leak a sandbox.
- Is this "if" statement necessary? Since the handler is set a few lines above
- I think it would be safer to make sure the handler is never null & get rid of if statement.

go-packages/meep-sessions/session-mgr.go
```diff
@@ -114,3 +125,56 @@ func (sm *SessionMgr) Authorizer(inner http.Handler) http.Handler {
 		}
 	})
 }
+
+// StartSessionWatchdog -
+func (sm *SessionMgr) StartSessionWatchdog(handler SessionTimeoutHandler) {
+	// Update Watchdog callback function
+	sm.wdHandler = handler
```
roymx, line 132: In relation to my next comment
- the handler should set inside the below `if !sm.wdStarted` statement - this way it's not possible to swap the handler at runtime. To swap user must stop and restart.
- if this function is called with a nil handler, we should simply return with an error before doing any further processing.

go-apps/meep-platform-ctrl/api/swagger.yaml
dilallkx: I also considered POSTing a heartbeat to a separate endpoint as an implementation option, therefore I will make the API modification as suggested.

js-packages/meep-platform-ctrl-client/test/api/UserAuthenticationApi.spec.js
dilallkx: This is unmodified, swagger-generated code.

go-apps/meep-platform-ctrl/server/user_authentication.go
dilallkx: Removed.

go-packages/meep-sessions/session-mgr.go
dilallkx: Code changes:
- StartSessionWatchdog() now returns an error code. Error cases:
  - Handler is nil
  - Watchdog already runnning
This guarantees that a running Session Watchdog will have a valid handler

roymx (approved):

PR #130, opened 2020-09-14
AdvantEDGE services handling of UE movement between different RAT (4G, 5G, Wifi)

pastorsx:
Support for the following events in Services mentioned below:

Location Service
1) - POA change using the name of the element. Valid for all poa types (4G, 5G, Wifi)

RNIS
2) - cell change only within poa-4G and poa-5G elements
3) - rab-establishment when going from poa-5G or Wifi to poa-4G
4) - rab-release when going to poa-5G or Wifi from poa-4G

5) PLMNInfo GET updated to only display the PLMN, no cellId

Tested using a scenario mixed with all types within different zones.

### DIFF
--- a/go-apps/meep-rnis/sbi/rnis-sbi.go
+++ b/go-apps/meep-rnis/sbi/rnis-sbi.go
@@ -216,9 +216,8 @@ func processActiveScenarioUpdate() {
 								cellId = poa.Poa5GConfig.CellId
 							}
 						}
-					case mod.NodeTypePoaWifi:
-						cellId = ""
 					default:
+						//empty cells for POAs not supporting RNIS
 						cellId = ""
 					}
 
@@ -288,9 +287,8 @@ func processActiveScenarioUpdate() {
 									cellId = nl.Poa5GConfig.CellId
 								}
 							}
-						case mod.NodeTypePoaWifi:
-							cellId = ""
 						default:
+							//empty cells for POAs not supporting RNIS
 							cellId = ""
 						}
 
--- a/go-apps/meep-rnis/sbi/rnis-sbi.go
+++ b/go-apps/meep-rnis/sbi/rnis-sbi.go
@@ -34,7 +34,7 @@ type SbiCfg struct {
 	RedisAddr      string
 	PostgisHost    string
 	PostgisPort    string
-	UeDataCb       func(string, string, string, string, int32)
+	UeDataCb       func(string, string, string, string, bool)
 	AppEcgiInfoCb  func(string, string, string, string)
 	ScenarioNameCb func(string)
 	CleanUpCb      func()
@@ -46,7 +46,7 @@ type RnisSbi struct {
 	handlerId            int
 	activeModel          *mod.Model
 	pc                   *postgis.Connector
-	updateUeDataCB       func(string, string, string, string, int32)
+	updateUeDataCB       func(string, string, string, string, bool)
 	updateAppEcgiInfoCB  func(string, string, string, string)
 	updateScenarioNameCB func(string)
 	cleanUpCB            func()
@@ -196,18 +196,31 @@ func processActiveScenarioUpdate() {
 					mnc := ""
 					mcc := ""
 					cellId := ""
+					erabIdValid := false
 					if domain.CellularDomainConfig != nil {
 						mnc = domain.CellularDomainConfig.Mnc
 						mcc = domain.CellularDomainConfig.Mcc
+						cellId = domain.CellularDomainConfig.DefaultCellId
 					}
-					if poa.Poa4GConfig != nil {
-						if poa.Poa4GConfig.CellId != "" {
-							cellId = poa.Poa4GConfig.CellId
-						} else {
-							cellId = domain.CellularDomainConfig.DefaultCellId
+					switch poa.Type_ {
+					case mod.NodeTypePoa4G:
+						if poa.Poa4GConfig != nil {
+							if poa.Poa4GConfig.CellId != "" {
+								cellId = poa.Poa4GConfig.CellId
+							}
 						}
+						erabIdValid = true
+					case mod.NodeTypePoa5G:
+						if poa.Poa5GConfig != nil {
+							if poa.Poa5GConfig.CellId != "" {
+								cellId = poa.Poa5GConfig.CellId
+							}
+						}
+					case mod.NodeTypePoaWifi:
+						cellId = ""
 					}
-					sbi.updateUeDataCB(name, mnc, mcc, cellId, -1)
+
+					sbi.updateUeDataCB(name, mnc, mcc, cellId, erabIdValid)
 				}
 			}
 		}
@@ -258,17 +271,23 @@ func processActiveScenarioUpdate() {
 						if domain.CellularDomainConfig != nil {
 							mnc = domain.CellularDomainConfig.Mnc
 							mcc = domain.CellularDomainConfig.Mcc
+							cellId = domain.CellularDomainConfig.DefaultCellId
 						}
-						if nl.Poa4GConfig != nil {
-							if nl.Poa4GConfig.CellId != "" {
-								cellId = nl.Poa4GConfig.CellId
-							} else {
-								cellId = domain.CellularDomainConfig.DefaultCellId
+						switch nl.Type_ {
+						case mod.NodeTypePoa4G:
+							if nl.Poa4GConfig != nil {
+								if nl.Poa4GConfig.CellId != "" {
+									cellId = nl.Poa4GConfig.CellId
+								}
 							}
-						} else {
-							if domain.CellularDomainConfig != nil {
-								cellId = domain.CellularDomainConfig.DefaultCellId
+						case mod.NodeTypePoa5G:
+							if nl.Poa5GConfig != nil {
+								if nl.Poa5GConfig.CellId != "" {
+									cellId = nl.Poa5GConfig.CellId
+								}
 							}
+						case mod.NodeTypePoaWifi:
+							cellId = ""
 						}
 
 						sbi.updateAppEcgiInfoCB(appName, mnc, mcc, cellId)
--- a/go-apps/meep-rnis/sbi/rnis-sbi.go
+++ b/go-apps/meep-rnis/sbi/rnis-sbi.go
@@ -218,6 +218,8 @@ func processActiveScenarioUpdate() {
 						}
 					case mod.NodeTypePoaWifi:
 						cellId = ""
+					default:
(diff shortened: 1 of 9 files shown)


go-apps/meep-rnis/sbi/rnis-sbi.go
```diff
@@ -175,18 +175,31 @@ func processActiveScenarioUpdate() {
...
-					if poa.Poa4GConfig != nil {
-						if poa.Poa4GConfig.CellId != "" {
-							cellId = poa.Poa4GConfig.CellId
-						} else {
-							cellId = domain.CellularDomainConfig.DefaultCellId
+					switch poa.Type_ {
+					case mod.NodeTypePoa4G:
+						if poa.Poa4GConfig != nil {
+							if poa.Poa4GConfig.CellId != "" {
+								cellId = poa.Poa4GConfig.CellId
```
roymx, line 188: What happens in the rest of the code if the scenario does not have a CellId set?
- should use what is set else set a well-known hardcoded default value (ex. 0000000 or FFFFFFF)
- same applies for 5G.

go-apps/meep-rnis/sbi/rnis-sbi.go
```diff
@@ -175,18 +175,31 @@ func processActiveScenarioUpdate() {
...
+							}
+						}
+						erabIdValid = true
+					case mod.NodeTypePoa5G:
+						if poa.Poa5GConfig != nil {
+							if poa.Poa5GConfig.CellId != "" {
+								cellId = poa.Poa5GConfig.CellId
+							}
 						}
+					case mod.NodeTypePoaWifi:
```
roymx, line 198: Use "default:" for the unsupported technology types. This way we will not have to re-work this if we add for say mmWave

go-apps/meep-rnis/sbi/rnis-sbi.go
```diff
@@ -175,18 +175,31 @@ func processActiveScenarioUpdate() {
...
 						mcc = domain.CellularDomainConfig.Mcc
+						cellId = domain.CellularDomainConfig.DefaultCellId
 					}
-					if poa.Poa4GConfig != nil {
-						if poa.Poa4GConfig.CellId != "" {
-							cellId = poa.Poa4GConfig.CellId
-						} else {
-							cellId = domain.CellularDomainConfig.DefaultCellId
+					switch poa.Type_ {
+					case mod.NodeTypePoa4G:
```
roymx, line 185: What happens in the rest of the code if the scenario does not have a CellId set?

should use what is set else set a well-known hardcoded default value (ex. 0000000 or FFFFFFF)
same applies for 5G.

go-apps/meep-rnis/sbi/rnis-sbi.go
```diff
@@ -230,19 +243,24 @@ func processActiveScenarioUpdate() {
...
-						} else {
-							if domain.CellularDomainConfig != nil {
-								cellId = domain.CellularDomainConfig.DefaultCellId
+						case mod.NodeTypePoa5G:
+							if nl.Poa5GConfig != nil {
+								if nl.Poa5GConfig.CellId != "" {
+									cellId = nl.Poa5GConfig.CellId
+								}
 							}
+						case mod.NodeTypePoaWifi:
```
roymx, line 261: Use "default:" for the unsupported technology types. This way we will not have to re-work this if we add for say mmWave

go-apps/meep-rnis/server/rnis.go
```diff
@@ -230,13 +230,13 @@ func updateUeData(name string, mnc string, mcc string, cellId string, erabId int
 	//updateDB if changes occur
 	if newEcgi.Plmn.Mnc != oldPlmnMnc || newEcgi.Plmn.Mcc != oldPlmnMcc || newEcgi.CellId != oldCellId {
 
-		//allocating a new erabId if entering a 3GPP environment
-		if oldCellId == "" && cellId != "" {
+		//allocating a new erabId if entering a 4G environment (existence of an erabId)
```
roymx, line 233: The logic below is confusing because there are two conditions checked in the if-else statement + the logic changes (i.e. == -1 vs != -1)

The pseudo-code should look like this

`
if oldErabId == -1{
   if erabIdValid {
      // do something
   }
   else {
      // do something else
   }
}
else{
   if erabIdValid {
      // do something
   }
   else {
      // do something else
   }

}
`

go-apps/meep-rnis/server/rnis.go
```diff
@@ -568,7 +566,7 @@ func checkReNotificationRegisteredSubscriptions(appId string, assocId *Associate
 				match = false
 			}
 
-			if match && (((sub.FilterCriteria.Plmn == nil) || (sub.FilterCriteria.Plmn != nil && (newPlmn != nil && newPlmn.Mnc == sub.FilterCriteria.Plmn.Mnc && newPlmn.Mcc == sub.FilterCriteria.Plmn.Mcc))) && (oldPlmn == nil || oldCellId == "")) {
+			if match && (((sub.FilterCriteria.Plmn == nil) || (sub.FilterCriteria.Plmn != nil && (newPlmn != nil && newPlmn.Mnc == sub.FilterCriteria.Plmn.Mnc && newPlmn.Mcc == sub.FilterCriteria.Plmn.Mcc))) && (oldPlmn == nil || oldCellId == "" || erabId != -1)) {
```
roymx, line 569: It is difficult to understand what is going-on in the if statement above... it would be a good idea to replace the whole block with this `match = myNewFunction(param1, param2, param3, etc.)` and expand the logic (adding a few comments)

go-apps/meep-rnis/server/rnis.go
```diff
@@ -665,7 +658,7 @@ func checkRrNotificationRegisteredSubscriptions(appId string, assocId *Associate
 				match = false
 			}
 
-			if match && (((sub.FilterCriteria.Plmn == nil) || (sub.FilterCriteria.Plmn != nil && (oldPlmn != nil && oldPlmn.Mnc == sub.FilterCriteria.Plmn.Mnc && oldPlmn.Mcc == sub.FilterCriteria.Plmn.Mcc))) && (newPlmn == nil || newCellId == "")) {
+			if match && (((sub.FilterCriteria.Plmn == nil) || (sub.FilterCriteria.Plmn != nil && (oldPlmn != nil && oldPlmn.Mnc == sub.FilterCriteria.Plmn.Mnc && oldPlmn.Mcc == sub.FilterCriteria.Plmn.Mcc))) && (newPlmn == nil || newCellId == "" || erabId == -1)) {
```
roymx, line 661: It is really difficult to understand what is going-on in the if statement above... it would be a good idea to replace the whole block with this `match = myNewFunction(param1, param2, param3, etc.)` and expand the logic (adding a few comments)

roymx (changes requested):
Please apply the requested changes

dilallkx:
Please rebase this branch to the tip of develop.

go-apps/meep-rnis/sbi/rnis-sbi.go
pastorsx: added a default cases that is neither 4g-5g or wifi.. so will apply to new technology or POA-GENERIC

go-apps/meep-rnis/sbi/rnis-sbi.go
pastorsx: cellId is set to a default value if a default value is provided in the operator cell element if it is not defined in the 4G or 5G poa config. If it is not set anywhere, it is not set and current services, like RNIS will not consider this as a valid poa (because a cellId is necessary). If cellId is not set, it is an empty string which is used when checking subscription and notification will be sent accordingly. No cellId implies the POA-GENERIC behaviour, not a 3gpp or wifi element so no support for the services.

go-apps/meep-rnis/sbi/rnis-sbi.go
pastorsx: default added

go-apps/meep-rnis/server/rnis.go
pastorsx: fixed using the pseudocode and added comments to be more explicit

go-apps/meep-rnis/server/rnis.go
pastorsx: i agree that this could be made simpler through using a function, but this code will need rework when supporting V2 because subscription changed and it has effects on the filters, so i suggest reworking it when implementing v2.

go-apps/meep-rnis/server/rnis.go
pastorsx: i agree that this could be made simpler through using a function, but this code will need rework when supporting V2 because subscription changed and it has effects on the filters, so i suggest reworking it when implementing v2.

go-apps/meep-rnis/sbi/rnis-sbi.go
roymx: These kind of implications makes the code hard to follow: "No cellId implies the POA-GENERIC behaviour" 
How is someone supposed to know that from reading the code?

pastorsx: its actually no cellId implies non 4G or 5G behavior, so everything else. Including Wifi but wifi has its own case, while generic POA falls into the default case....

go-apps/meep-rnis/sbi/rnis-sbi.go
roymx: - Remove `case PoaWifi` as it does the same as `default` + RNIS should only care about 3GPP tech.
- Add this comment in default: `// Empty cellId is ignored by RNIS`

go-apps/meep-rnis/sbi/rnis-sbi.go
roymx: - Remove `case PoaWifi` as it does the same as `default` + RNIS should only care about 3GPP tech.
- Add this comment in default: `// Empty cellId is ignored by RNIS`

go-apps/meep-rnis/server/rnis.go
roymx: I insist, please clarify this section of code as we may not support V2 for some time.

go-apps/meep-rnis/server/rnis.go
roymx: I insist, please clarify this section of code as we may not support V2 for some time.

roymx (changes requested):
Some review comments were not properly understood; please update accordingly.

roymx (approved):

PR #138, opened 2020-10-02
Configurable Maximum Session Count

dilallkx:
**CHANGES:**
- Platform Controller:
  - MEEP_MAX_SESSIONS environment variable to configure max simultaneous session count
- Session Store:
  - New method to retrieve current session count
- loc-serv, rnis, wais:
  - Flush scenario data on service startup
- permissions.yaml
  - Sample endpoint permissions for WAIS

**TESTING:**
- Cypress & UT pass
- Manual verification of max session count

### DIFF
--- a/charts/meep-platform-ctrl/values.yaml
+++ b/charts/meep-platform-ctrl/values.yaml
@@ -25,6 +25,7 @@ image:
   pullPolicy: Always
   env:
     MEEP_SESSION_KEY: "my-secret-key"
+    MEEP_MAX_SESSIONS: "10"
 
 service:
   type: ClusterIP
--- a/go-apps/meep-platform-ctrl/server/user_authentication.go
+++ b/go-apps/meep-platform-ctrl/server/user_authentication.go
@@ -59,7 +59,7 @@ func uaLoginUser(w http.ResponseWriter, r *http.Request) {
 		if count >= pfmCtrl.maxSessions {
 			err = errors.New("Maximum session count exceeded")
 			log.Error(err.Error())
-			http.Error(w, err.Error(), http.StatusInsufficientStorage)
+			http.Error(w, err.Error(), http.StatusServiceUnavailable)
 			return
 		}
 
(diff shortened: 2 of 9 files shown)


charts/meep-platform-ctrl/values.yaml
```diff
@@ -25,6 +25,7 @@ image:
   pullPolicy: Always
   env:
     MEEP_SESSION_KEY: "my-secret-key"
+    MEEP_MAX_SESSIONS: "10"
```
roymx, line 28: I guess this is OK to specify this limit here for now.
This value should be deployment specific and therefore defined elsewhere in the deployment configuration.

go-apps/meep-platform-ctrl/server/user_authentication.go
```diff
@@ -54,6 +54,15 @@ func uaLoginUser(w http.ResponseWriter, r *http.Request) {
 	sessionStore := pfmCtrl.sessionMgr.GetSessionStore()
 	session, err := sessionStore.GetByName(username)
 	if err != nil {
+		// Check if max session count is reached before creating a new one
+		count := sessionStore.GetCount()
+		if count >= pfmCtrl.maxSessions {
+			err = errors.New("Maximum session count exceeded")
+			log.Error(err.Error())
+			http.Error(w, err.Error(), http.StatusInsufficientStorage)
```
roymx, line 62: Error returned should be 503 Service Unavailable
https://www.restapitutorial.com/httpstatuscodes.html
503 Service Unavailable
The server is currently unable to handle the request **due to a temporary overloading** or maintenance of the server. The implication is that this is a temporary condition which will be alleviated after some delay. If known, the length of the delay MAY be indicated in a Retry-After header. If no Retry-After is given, the client SHOULD handle the response as it would for a 500 response.

roymx (changes requested):
Only a return code needs to be changed

charts/meep-platform-ctrl/values.yaml
dilallkx: This is the default value. It can/should be overridden @ deployment, just like the session key. This is done in the user/values override yaml file.

go-apps/meep-platform-ctrl/server/user_authentication.go
dilallkx: Updated as suggested.

roymx (approved):

PR #141, opened 2020-10-05
Meepctl Helm fix + Version bump to 1.5.4

dilallkx:
**CHANGES:**
- Bumped version to 1.5.4 to force a meepctl update
  - Required for helm v3 update
- Fixed helm output parsing error due to possible "Warning" output with helm v3.3+

**TESTING:**
- Cypress & UT pass

### DIFF
--- a/go-apps/meepctl/utils/helm.go
+++ b/go-apps/meepctl/utils/helm.go
@@ -37,7 +37,7 @@ func IsHelmRelease(name string, cobraCmd *cobra.Command) (exist bool, err error)
 	if verbose {
 		fmt.Println("Cmd:", cmd.Args)
 	}
-	out, err := cmd.CombinedOutput()
+	out, err := cmd.Output()
 	elapsed := time.Since(start)
 	if err != nil {
 		err = errors.New("Error listing component [" + name + "]")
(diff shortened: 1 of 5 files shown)


nikhildoifode (approved):

go-apps/meepctl/utils/helm.go
```diff
@@ -37,15 +37,14 @@ func IsHelmRelease(name string, cobraCmd *cobra.Command) (exist bool, err error)
 	if verbose {
 		fmt.Println("Cmd:", cmd.Args)
 	}
-	out, err := cmd.CombinedOutput()
+	out, err := cmd.Output()
```
roymx, line 40: As discussed, CombinedOutput provides error and warnings that are necessary for the verbose.
I would recommend to parse the CombinedOutput to see if there is (1) a line that starts with the name and (2) that line is an exact match to the name (this excludes release that may start with the same name)

roymx (approved):

go-apps/meepctl/utils/helm.go
dilallkx: Modified code to parse combined output and search for a line with an exact match, which is what this function should do. A function to check for a matching prefix can be added when required.

PR #144, opened 2020-10-21
Added playbooks for installation of NPM, Node, and GoLang

nikhildoifode:
Changes:
- Added code for installation of NPM, Node, and GoLang, ESLint and Golanci-lint
- Added current version check for NPM, Node and Golang

Test:
- Manually checked the installation of above things in local environment

### DIFF
--- a/playbooks/install-development-env.yml
+++ b/playbooks/install-development-env.yml
@@ -1,0 +1,13 @@
+---
+
+- hosts: cluster
+  gather_facts: yes
+  become: yes
+  roles:
+    - { role: golang, tags: golang }
+
+- hosts: cluster
+  gather_facts: yes
+  become: yes
+  roles:
+    - { role: node, tags: node }
--- a/playbooks/install-development-env.yml
+++ b/playbooks/install-development-env.yml
@@ -1,12 +1,12 @@
 ---
 
-- hosts: cluster
+- hosts: dev
   gather_facts: yes
   become: yes
   roles:
     - { role: golang, tags: golang }
 
-- hosts: cluster
+- hosts: dev
   gather_facts: yes
   become: yes
   roles:
--- a/playbooks/roles/node/tasks/install.yml
+++ b/playbooks/roles/node/tasks/install.yml
@@ -1,0 +1,31 @@
+---
+
+- name: Install required system packages
+  apt:
+    name: "{{ item }}"
+    state: present
+    update_cache: yes
+  with_items:
+    - build-essential
+    - libssl-dev
+
+- name: "Copy nvm file"
+  copy:
+    src: "install_nvm.sh"
+    dest: "install_nvm.sh"
+    mode: 0744
+  tags: node
+
+- name: "Install nvm"
+  shell: "bash install_nvm.sh"
+  tags: node
+  args:
+    executable: /bin/bash
+
+- name: Install node
+  become: false
+  shell: /bin/bash -c "source /home/{{ ansible_user }}/.nvm/nvm.sh && nvm ls-remote | grep 'Latest LTS'; nvm install {{ node_version }}"
+
+- name: Install npm and eslint
+  become: false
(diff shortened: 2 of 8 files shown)


roymx (approved):

playbooks/install-development-env.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+- hosts: cluster
```
dilallkx, line 3: We should add a [dev] target in hosts.ini that specifies the development machine(s) where the dev environment has to be installed. We don't need to install go, node & npm on all cluster nodes.

playbooks/roles/node/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+  failed_when: "'command not found' in npm_check.stdout"
+
+- debug:
+    msg: "[WARNING] Current npm version is lower than recommended version {{ npm_version }}"
+  when: npm_check.stdout != '' and npm_check.stdout is version_compare(npm_version, '<')
+
+- name: Install Node and NPM if node or npm command not found
+  include_tasks:
+    file: install.yml
+  when: node_check.stdout == '' or npm_check.stdout == ''
```
dilallkx, line 31: If an older version of node is installed but no npm is installed, this will invoke the install script, correct? We should only install if neither is already installed.

playbooks/roles/node/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+      - ansible_distribution_version == '18.04'
+    quiet: yes
+
+- name: Check if node is installed
+  command: /bin/bash -c "source /home/{{ ansible_user }}/.nvm/nvm.sh; node -v | cut -c 2-"
+  register: node_check
+  failed_when: "'command not found' in node_check.stdout"
+
+- debug:
+    msg: "[WARNING] Current node version is lower than recommended version {{ node_version }}"
```
dilallkx, line 16: We should tell the user what to do to get the script to install properly. Example: tell user to uninstall existing version and rerun script, an also to remove the .nvm folder, etc...

dilallkx: NOTE: Same comment applies for other warning messages as well, unless we provide an uninstall playbook.

playbooks/roles/node/tasks/install.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: "Install nvm"
+  become: false
+  shell: "bash install_nvm.sh"
+  tags: node
+  args:
+    executable: /bin/bash
+
+- name: Install node
+  become: false
+  shell: /bin/bash -c "source /home/{{ ansible_user }}/.nvm/nvm.sh && nvm ls-remote | grep 'Latest LTS'; nvm install {{ node_version }}"
```
dilallkx, line 28: Is this necessary: 'nvm ls-remote | grep 'Latest LTS''?

dilallkx (changes requested):

playbooks/roles/node/tasks/install.yml
nikhildoifode: I followed the wiki for this. So need to check whether it is required in the wiki itself or not.
https://github.com/InterDigitalInc/AdvantEDGE/wiki/development-environment#step-4---install-nodejs-latest-lts-version

dilallkx: In the manual procedure it is required to retrieve the latest version to install. Since we set the version to install we should not have to list the currently available versions.

playbooks/install-development-env.yml
nikhildoifode: Added dev target

playbooks/roles/node/tasks/main.yml
nikhildoifode: Changed to and condition

playbooks/roles/node/tasks/main.yml
nikhildoifode: Added extra steps to where ever warning was showed.

playbooks/roles/node/tasks/install.yml
nikhildoifode: Ok removed

dilallkx (approved):

PR #151, opened 2020-11-10
Session & OAuth configuration with k8s secrets

dilallkx:
**CHANGES:**
- Deployment config:
  - Bumped version to 1.5.8
  - Replaced clear text OAuth & session keys by references to k8s secrets
- Core service charts:
  - Added support for environment variables provisioned from k8s secrets
- meepctl:
  - Bumped version to 1.5.8
  - Updated code to override secrets using deployment config
- New secret configuration script to provision k8s secrets from file
  - NOTE: Once provisioned in a secret, the secrets.yaml file can/should be removed from the host machine
- Bumped demo1 & demo2 versions to match backend model version
- Session Manager:
  - Added sandbox name verification during endpoint authorization
    - Normal users can no longer use their session cookie to access other sandbox endpoints
    - Admin users still have full access to all sandboxes
- Virt Engine:
  - Moved generated charts from persistent volume to pod filesystem
    - Charts were in PV for debugging purposes initially (and eventually for persistence on pod restart which is not used yet)
    - Because the Virt Engine creates user charts for sandbox microservices and stores the session key in the chart, I removed them from the PV so that they could no longer be accessed directly on the host filesystem

**TESTING:**
- Cypress & UT pass
- Manual verification of deployment with & without session & oauth secrets


config/configure-secrets.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/python
```
roymx, line 1: What's the rationale for adding a dependency on Python?

roymx (approved):

config/configure-secrets.py
dilallkx: Simpler yaml file parsing modules.

PR #160, opened 2020-11-26
Cypress for AdvantEDGE release 1.6

pastorsx:
cypress test updated to validate every elements in frontend added in v1.6 or before
- configuration
- execution  
  - pane visibility (event - dashboard)
  - event button (automation, replay, save replay)
 
cypress test: PASS

### DIFF
--- a/js-apps/meep-frontend/src/js/containers/cfg/cfg-page-container.js
+++ b/js-apps/meep-frontend/src/js/containers/cfg/cfg-page-container.js
@@ -596,6 +596,7 @@ class CfgPageContainer extends Component {
           title="Open Scenario"
           open={this.props.currentDialog === IDC_DIALOG_OPEN_SCENARIO}
           options={this.props.scenarios}
+          onDismiss={() => { this.closeDialog();}}
           onClose={() => {
             this.closeDialog();
           }}
--- a/js-apps/meep-frontend/src/js/containers/cfg/cfg-page-container.js
+++ b/js-apps/meep-frontend/src/js/containers/cfg/cfg-page-container.js
@@ -596,7 +596,6 @@ class CfgPageContainer extends Component {
           title="Open Scenario"
           open={this.props.currentDialog === IDC_DIALOG_OPEN_SCENARIO}
           options={this.props.scenarios}
-          onDismiss={() => { this.closeDialog();}}
           onClose={() => {
             this.closeDialog();
           }}
--- a/test/cypress/integration/tests/scenario-exec-spec.js
+++ b/test/cypress/integration/tests/scenario-exec-spec.js
@@ -55,7 +55,7 @@ describe('Scenario Execution', function () {
 
   // Test Variables
   let defaultScenario = 'None';
-  let sandbox = 'sbox-test3';
+  let sandbox = 'sbox-test';
   let scenario = 'demo1';
   let scenario2 = 'demo2';
   let replayEventsName = 'replaydemo1'
--- a/test/cypress/integration/tests/scenario-exec-spec.js
+++ b/test/cypress/integration/tests/scenario-exec-spec.js
@@ -48,7 +48,7 @@ import {
 } from '../../../../js-apps/meep-frontend/src/js/util/elem-utils';
 
 // Import Test utility functions
-import { selector, click, type, select, verify, verifyEnabled, verifyForm } from '../util/util';
+import { selector, click, check, type, select, verify, verifyEnabled, verifyForm, verifyChecked } from '../util/util';
 
 // Scenario Execution Tests
 describe('Scenario Execution', function () {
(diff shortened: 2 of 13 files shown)


test/cypress/integration/tests/scenario-exec-spec.js
```diff
@@ -273,6 +369,137 @@ describe('Scenario Execution', function () {
...
+    } else {
+      check(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_CHKBOX_LOOP, false);
+    }
+    select(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_EVT_REPLAY_FILES, name);
+    cy.wait(500);
+    verifyEnabled(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_BTN_REPLAY_START, true);
+    click(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_BTN_REPLAY_START)
+    //check right number of events will be executed
+    cy.contains("EVENT COUNT");
+    cy.contains("/ 13");
```
dilallkx, line 430: It is difficult to maintain this. If we update the mobility events then this number must be changed.

To fix this we would need to restart the scenario and send a smaller number of events (e.g. 1 mobility + 1 net char). We would then always have these 2 events only. However, the scenario teardown & bring-up is what takes the most time, which would make the overall test longer.

Recommendation: leave as is for now, but add comments in code to clearly state where this event count is coming from and how to calculate it.

test/cypress/integration/tests/scenario-exec-spec.js
```diff
@@ -273,6 +369,137 @@ describe('Scenario Execution', function () {
...
+    verifyEnabled(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_BTN_REPLAY_START, false);
+    cy.wait(75000);
+    //confirm it executed completely
+    verifyEnabled(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_BTN_REPLAY_START, true);
+
+    testAutoReplayExecution(name, true)
+    cy.wait(1000);
+    //confirm execution started
+    verifyEnabled(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_BTN_REPLAY_START, false);
+    cy.wait(75000);
```
dilallkx, line 407: Does the actual replay time take close to 75 seconds?
Can this timer be reduced?

test/cypress/integration/tests/scenario-exec-spec.js
dilallkx: Also, I believe the 'verify()' call can be used to search for text contained within a specific element. We should try to verify that only the status bar contains this text.

dilallkx (changes requested):

js-apps/meep-frontend/src/js/containers/cfg/cfg-page-container.js
```diff
@@ -596,6 +596,7 @@ class CfgPageContainer extends Component {
           title="Open Scenario"
           open={this.props.currentDialog === IDC_DIALOG_OPEN_SCENARIO}
           options={this.props.scenarios}
+          onDismiss={() => { this.closeDialog();}}
```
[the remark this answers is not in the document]
pastorsx: was meant to fix the click outside the window... but it was only working using ESC, not clicking.. and i forgot it there... removed

test/cypress/integration/tests/scenario-exec-spec.js
```diff
@@ -48,44 +48,117 @@ import {
...
 
 // Scenario Execution Tests
 describe('Scenario Execution', function () {
 
   // Test Variables
   let defaultScenario = 'None';
   let sandbox = 'sbox-test';
-
-  // Test Setup
-  beforeEach(() => {
```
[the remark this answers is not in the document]
pastorsx: the beforeEach doesn;t seem to always work. So some test cases sometimes fail and we need to run cypress again because it tried to execute the test case but somehow didn't run the portion in the before each. The content of the beforeEach was moved to a function that is invoked every time now... and cypress always work as expected (otherwise, trying to click some buttons not already enabled)...

test/cypress/integration/tests/scenario-exec-spec.js
```diff
@@ -153,7 +249,7 @@ describe('Scenario Execution', function () {
     cy.wait(1000);
     click(meep.EXEC_BTN_DELETE_SANDBOX);
     click(meep.MEEP_DLG_DELETE_SANDBOX, 'Ok');
-    cy.wait(10000);
+    cy.wait(20000);
```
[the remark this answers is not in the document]
pastorsx: simply because deletion of the sandbox could take up to 15 seconds and that was failing the next test case because it was not ready yet,,,

test/cypress/integration/tests/scenario-exec-spec.js
```diff
@@ -273,6 +369,137 @@ describe('Scenario Execution', function () {
...
+  function testAutoReplay(name) {
+    cy.log('Auto Replay');
+
+    testAutoReplayExecution(name, false)
+    cy.wait(1000);
+    //confirm execution started
+    verifyEnabled(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_BTN_REPLAY_START, false);
+    cy.wait(75000);
+    //confirm it executed completely
+    verifyEnabled(meep.EXEC_BTN_EVENT_BTN_AUTO_REPLAY_BTN_REPLAY_START, true);
```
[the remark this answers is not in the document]
pastorsx: fixed

test/cypress/integration/tests/scenario-exec-spec.js
pastorsx: used in conjonction with verifyEnabled... max should then be 75 sec.. could be quicker...

pastorsx: verify enabled now used where the replay is executing once... otherwise, when in loop, still waiting for the whole 75s

test/cypress/integration/tests/scenario-exec-spec.js
pastorsx: fixed... added comments and also number of events now passed in param to the function.. so it is clear.. also added a label to check rather than whole page

dilallkx (approved):

PR #182, opened 2021-01-13
Scenario Name check in platform API

pastorsx:
Scenario name MUST be identical both in the endpoint and in the body content otherwise an error is sent.

Otherwise, creating multiple entries (based on the endpoint name) that points to the same scenario name is possible.

### DIFF
--- a/go-packages/meep-model/validator.go
+++ b/go-packages/meep-model/validator.go
@@ -59,7 +59,7 @@ func createNetChar(lat int32, latVar int32, dist string, tputDl int32, tputUl in
 }
 
 // ValidateScenario - Verify if json scenario is valid & supported. Upgrade scenario if possible & necessary.
-func ValidateScenario(jsonScenario []byte) (validJsonScenario []byte, status string, err error) {
+func ValidateScenario(jsonScenario []byte, name string) (validJsonScenario []byte, status string, err error) {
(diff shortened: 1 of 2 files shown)


go-packages/meep-model/validator.go
```diff
@@ -70,6 +70,14 @@ func ValidateScenario(jsonScenario []byte) (validJsonScenario []byte, status str
...
 		return nil, ValidatorStatusError, err
 	}
+
+	if name != "" {
+		if scenario.Name != name {
+                        err = errors.New("Scenario creation name " + name + " incompatible with scenario body content name " + scenario.Name + ". They must be the same.")
+                        return nil, ValidatorStatusError, err
+		}
+	}
+
```
dilallkx, line 80: This code should be in the private _validateScenario_ function where the scenario format is verified.

dilallkx (changes requested):

go-packages/meep-model/validator.go
pastorsx: After discussions, leaving it as is. Versions is already checked at that level and using the private validateScenario function more as a validate deployment. Checks in the private function is limited to name and id uniqueness.

dilallkx (approved):

PR #200, opened 2021-02-09
L2Meas endpoint implementation

pastorsx:
Implementation of the content of the L2Meas

using query parameters such as:
- app_instance_id
- cellId
- ipaddress field (name of the ue based on model)

implementation of the content:
cellInfo
- number of UEs (nongbr)
- delay (air interface)
- packet loss (air interface)

cellUEInfo
- delay (air interface only)
- packet loss (some of all apps, so calculated) integer 0..100 base on packet loss between 10^-4 to 10^-6 as per spec
- throuput (calculated)
- dataVolume (calculated)

Note: packet loss is not so representative (pretty baniry at the moment, either 0 or 100) because packet loss is minimum 0.01 ( value in the hundredth, rather than micro as per spec)... this limitation needs to be lifted. Tracked by another issue. 

Implementation explanation:
- RNIS now connectes to two different REDIS DB (rnis data itself populated by the sbi and table 0 where data collected from sidecar is stored)
- added more info in the redis rnis table for quicker access and more info such as the apps running on the ue so that it gets the data collected more easily (data collected based on meApp by sidecar, no knowledge of UE it belongs to)

testing:
ut passed

### DIFF
--- a/go-apps/meep-rnis/sbi/rnis-sbi.go
+++ b/go-apps/meep-rnis/sbi/rnis-sbi.go
@@ -326,7 +326,20 @@ func processActiveScenarioUpdate() {
 						throughputUL = ue.NetChar.ThroughputUl
 					}
 
-					sbi.updateUeDataCB(UeDataSbi{name, mnc, mcc, cellId, nrcellId, erabIdValid, appNames, latency, throughputUL, throughputDL, ploss})
+					var ueDataSbi = UeDataSbi{
+						Name:         name,
+						Mnc:          mnc,
+						Mcc:          mcc,
+						CellId:       cellId,
+						NrCellId:     nrcellId,
+						ErabIdValid:  erabIdValid,
+						AppNames:     appNames,
+						Latency:      latency,
+						ThroughputUL: throughputUL,
+						ThroughputDL: throughputDL,
+						PacketLoss:   ploss,
+					}
+					sbi.updateUeDataCB(ueDataSbi)
 				}
 			}
 		}
@@ -342,7 +355,12 @@ func processActiveScenarioUpdate() {
 			}
 		}
 		if !found {
-			sbi.updateUeDataCB(UeDataSbi{prevUeName, "", "", "", "", false, nil, 0, 0, 0, 0.0})
+			var ueDataSbi = UeDataSbi{
+				Name:        prevUeName,
+				ErabIdValid: false,
+			}
+
+			sbi.updateUeDataCB(ueDataSbi)
 			log.Info("Ue removed : ", prevUeName)
 		}
 	}
@@ -374,7 +392,16 @@ func processActiveScenarioUpdate() {
 				throughputUL = pl.NetChar.ThroughputUl
 			}
 
-			sbi.updateAppInfoCB(AppInfoSbi{appName, pl.Type_, pl.Name, latency, throughputUL, throughputDL, ploss})
+			var appInfoSbi = AppInfoSbi{
+				Name:         appName,
+				ParentType:   pl.Type_,
+				ParentName:   pl.Name,
+				Latency:      latency,
+				ThroughputUL: throughputUL,
+				ThroughputDL: throughputDL,
+				PacketLoss:   ploss,
+			}
+			sbi.updateAppInfoCB(appInfoSbi)
 		}
 	}
 
@@ -388,7 +415,11 @@ func processActiveScenarioUpdate() {
 			}
 		}
 		if !found {
-			sbi.updateAppInfoCB(AppInfoSbi{prevApp, "", "", 0, 0, 0, 0.0})
+			var appInfoSbi = AppInfoSbi{
+				Name: prevApp,
+			}
+
+			sbi.updateAppInfoCB(appInfoSbi)
 			log.Info("App removed : ", prevApp)
 		}
 	}
@@ -434,7 +465,18 @@ func processActiveScenarioUpdate() {
 					throughputUL = nl.NetChar.ThroughputUl
 				}
 
-				sbi.updatePoaInfoCB(PoaInfoSbi{name, nl.Type_, mnc, mcc, cellId, latency, throughputUL, throughputDL, ploss})
+				var poaInfoSbi = PoaInfoSbi{
+					Name:         name,
+					PoaType:      nl.Type_,
+					Mnc:          mnc,
+					Mcc:          mcc,
+					CellId:       cellId,
+					Latency:      latency,
+					ThroughputUL: throughputUL,
+					ThroughputDL: throughputDL,
+					PacketLoss:   ploss,
+				}
+				sbi.updatePoaInfoCB(poaInfoSbi)
 			}
 		}
 	}
--- a/go-apps/meep-rnis/sbi/rnis-sbi.go
+++ b/go-apps/meep-rnis/sbi/rnis-sbi.go
@@ -31,10 +31,10 @@ const moduleName string = "meep-rnis-sbi"
 type SbiCfg struct {
 	SandboxName    string
 	RedisAddr      string
-	UeDataCb       func(string, string, string, string, string, bool, []string)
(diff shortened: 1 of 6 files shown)


go-apps/meep-rnis/server/rnis.go
```diff
@@ -75,12 +76,16 @@ const MEAS_REP_UE_NOTIFICATION = "MeasRepUeNotification"
 const NR_MEAS_REP_UE_NOTIFICATION = "NrMeasRepUeNotification"
 
 var RNIS_DB = 5
+var DATA_DB = 0
+var RNIS_DB_CONNECTOR_INDEX = 0
+var DATA_DB_CONNECTOR_INDEX = 1
 
-var rc *redis.Connector
+var rc [2]*redis.Connector
```
dilallkx, line 83: Instead of using a redis connector directly, use the meep-metric-store package to access the network metrics. You can then revert all the changes make to create a slice of connectors.

go-apps/meep-rnis/server/rnis.go
```diff
@@ -2746,9 +2798,128 @@ func populateL2Meas(key string, jsonInfo string, l2MeasData interface{}) error {
...
+	for _, appName := range ueData.AppNames {
+
+		//we calculate stats for the queried app only or for all if none provided
+		if appName != data.queryAppInsId && data.queryAppInsId != "" {
+			continue
+		}
+		keyName := msBaseKey + "*" + appName
+		appStats := SumAppInfo{appName, 0, 0, 0, 0}
+
+		err = rc[DATA_DB_CONNECTOR_INDEX].ForEachEntry(keyName, calculateSum, &appStats)
```
dilallkx, line 2832: Replace with a call to meep-metrics-store: `GetCachedNetworkMetric("*", appName)`

dilallkx (changes requested):

go-apps/meep-rnis/server/rnis.go
```diff
@@ -2520,6 +2569,7 @@ func layer2MeasInfoGet(w http.ResponseWriter, r *http.Request) {
 	q := u.Query()
 	//meAppName := q.Get("app_ins_id")
 
+	l2MeasData.queryAppInsId = q.Get("app_ins_id")
```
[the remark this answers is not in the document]
pastorsx: others are arrays while this one is only one instance
q.Get() gets one instance only while q[] would be the array stored directly

go-apps/meep-rnis/server/rnis.go
```diff
@@ -2746,9 +2798,128 @@ func populateL2Meas(key string, jsonInfo string, l2MeasData interface{}) error {
...
+	data.l2Meas.CellInfo[cellIndex].DlNongbrPdrCell = poaPacketLoss
+	data.l2Meas.CellInfo[cellIndex].UlNongbrPdrCell = poaPacketLoss
+
+	//update ueInfo delay
+	//delay is the latency between air interface (POA<->UE)
+	data.l2Meas.CellUEInfo[cellUeIndex].DlNongbrDelayUe = latency //latency from the air interface only (POA)
+	data.l2Meas.CellUEInfo[cellUeIndex].UlNongbrDelayUe = latency
+	data.l2Meas.CellUEInfo[cellUeIndex].DlNongbrDataVolumeUe = ueStats.DlTraffic / 1000 //kbits
+	data.l2Meas.CellUEInfo[cellUeIndex].UlNongbrDataVolumeUe = ueStats.UlTraffic / 1000 //kbits
+	data.l2Meas.CellUEInfo[cellUeIndex].DlNongbrThroughputUe = ueStats.DlTraffic / 1000 //kbits/s
```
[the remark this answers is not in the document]
pastorsx: yes, data volume is the amount of data in kbits... throughput is the same value but in kbits/s... Out data volume interval calculation is considered to be one second, so it end up being the same as throughput. We do not set a variable for that interval and the spec is not mentioning anything about it either

go-apps/meep-rnis/server/rnis.go
```diff
@@ -2746,9 +2798,128 @@ func populateL2Meas(key string, jsonInfo string, l2MeasData interface{}) error {
...
+	jsonPoaData, _ := rc[RNIS_DB_CONNECTOR_INDEX].JSONGetEntry(baseKey+"POA:"+ueData.ParentPoaName, ".")
+
+	latency := int32(0)
+	poaPacketLoss := int32(0)
+	if jsonPoaData != "" {
+		poaDataObj := convertJsonToPoaInfo(jsonPoaData)
+		if poaDataObj != nil {
+			latency = poaDataObj.Latency
+			ploss := poaDataObj.PacketLoss
+			//return between 10^-4 t 10^-6
```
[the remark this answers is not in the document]
pastorsx: yes 0 is 10^-6 and 100 is 10^-4... which is why we currently always have 100 since the granularity stops at 0.01. QCIs are mainly between 10^-6 and 10^-4, which i guess this is why they used that, but the 0..100 range is limitative. This is from the spec.

go-apps/meep-rnis/server/rnis.go
```diff
@@ -2584,7 +2634,7 @@ func layer2MeasInfoGet(w http.ResponseWriter, r *http.Request) {
 }
 
 func populateL2MeasPOA(key string, jsonInfo string, l2MeasData interface{}) error {
-	// Get query params & userlist from user data
+	// et query params & userlist from user data
```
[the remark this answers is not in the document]
pastorsx: fixed

go-apps/meep-rnis/sbi/rnis-sbi.go
```diff
@@ -31,10 +31,10 @@ const moduleName string = "meep-rnis-sbi"
 type SbiCfg struct {
 	SandboxName    string
 	RedisAddr      string
-	UeDataCb       func(string, string, string, string, string, bool, []string)
+	UeDataCb       func(string, string, string, string, string, bool, []string, int32, int32, int32, float64)
```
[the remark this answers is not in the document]
pastorsx: i tried to avoid passing structures and wanted to use basic types... but will do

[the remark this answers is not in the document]
pastorsx: updated

go-apps/meep-rnis/sbi/rnis-sbi.go
```diff
@@ -269,14 +305,28 @@ func processActiveScenarioUpdate() {
...
+					throughputDL := int32(0)
+					throughputUL := int32(0)
+					if ue.NetChar != nil {
+						latency = ue.NetChar.Latency
+						ploss = ue.NetChar.PacketLoss
+						throughputDL = ue.NetChar.ThroughputDl
+						throughputUL = ue.NetChar.ThroughputUl
+					}
+
+					sbi.updateUeDataCB(UeDataSbi{name, mnc, mcc, cellId, nrcellId, erabIdValid, appNames, latency, throughputUL, throughputDL, ploss})
```
dilallkx, line 329: The purpose of adding the structure was to avoid argument ordering issues. By initializing the structure this way you will continue to have ordering issues. I recommend creating an instance and populating the fields by name directly above instead of creating temporary variables.

dilallkx: NOTE: Same comment for the other structures.

dilallkx (changes requested):

go-apps/meep-rnis/server/rnis.go
pastorsx: fixed

go-apps/meep-rnis/server/rnis.go
pastorsx: fixed

go-apps/meep-rnis/sbi/rnis-sbi.go
pastorsx: fixed on all 3 structures

dilallkx (approved):

PR #205, opened 2021-02-19
Packet loss decimal supports up to 6 decimals

pastorsx:
Packet loss was limited to 2 decimals although the frontend would allow 6 decimals. Issue was resolved to support the max limit of decimals supported by linux tc by using a string format.

test:
Executed and passed

NOTE: change to the frontend added as part of this fix in order to support netchar modification on the Terminal


dilallkx:
PR moved to fork.

PR #213, opened 2021-02-26
Virt Engine process lifecycle support

dilallkx:
**CHANGES:**
- meep-model:
  - Added event type & user data support to scenario update callback
  - Fixed UT & model package calls to support changes
- Sandbox Controller:
  - Sends scenario update message on global message queue for Virt Engine
  - Includes sandbox name, event type (Add/Modify/Remove node) & node name in message payload
- Virt Engine:
  - Created function to generate ME Svc charts
    - Renamed ME Svc charts to "me-svc" + <me svc name> instead of a mangled process-specific name
    - ME Services now always created in a separate chart (except egress services that must be improved)
  - Added scenario update message handler
    - Supports add, modify & remove for physical locations & processes 
    - Generates & removes the necessary charts + installs & uninstalls helm releases

**TESTING:**
- UT, System & Cypress test pass
- Manual testing of physical location & process add/modify/remove

### DIFF
--- a/go-apps/meep-sandbox-ctrl/server/sandbox-ctrl.go
+++ b/go-apps/meep-sandbox-ctrl/server/sandbox-ctrl.go
@@ -65,6 +65,8 @@ const serviceName = "Sandbox Controller"
 // MQ payload fields
 const fieldSandboxName = "sandbox-name"
 const fieldScenarioName = "scenario-name"
+const fieldEventType = "event-type"
+const fieldNodeName = "node-name"
 
 // Event types
 const (
@@ -516,7 +518,7 @@ func ceTerminateScenario(w http.ResponseWriter, r *http.Request) {
 	// Send Terminate message to Virt Engine on Global Message Queue
 	msg = sbxCtrl.mqGlobal.CreateMsg(mq.MsgScenarioTerminate, mq.TargetAll, mq.TargetAll)
 	msg.Payload[fieldSandboxName] = sbxCtrl.sandboxName
-	msg.Payload[fieldScenarioName] = ""
+	msg.Payload[fieldScenarioName] = sbxCtrl.activeModel.GetScenarioName()
 	log.Debug("TX MSG: ", mq.PrintMsg(msg))
 	err = sbxCtrl.mqGlobal.SendMsg(msg)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


go-apps/meep-sandbox-ctrl/server/sandbox-ctrl.go
```diff
@@ -767,11 +772,16 @@ func sendEventScenarioUpdate(event dataModel.Event) (error, int, string) {
 // Retrieve element name from type-specific structure
 func getScenarioNodeName(node *dataModel.ScenarioNode) string {
 	name := ""
-	if node.Type_ == mod.NodeTypeUE {
+	if mod.IsPhyLoc(node.Type_) {
```
roymx, line 775: `if node.Type_ == mod.NodeTypeUE {` became `if mod.IsPhyLoc(node.Type_) {`
This change does not seem equivalent to me as a UE is a Physical Location but not all Physical locations are UEs.
Please clarify...

go-apps/meep-virt-engine/server/virt-engine.go
```diff
@@ -175,6 +176,21 @@ func msgHandler(msg *mq.Msg, userData interface{}) {
...
 		activateScenario(msg.Payload[fieldSandboxName])
+	case mq.MsgScenarioUpdate:
+		log.Debug("RX MSG: ", mq.PrintMsg(msg))
+		eventType := msg.Payload[fieldEventType]
+		sandboxName := msg.Payload[fieldSandboxName]
+		nodeName := msg.Payload[fieldNodeName]
+		switch eventType {
+		case mod.EventAddNode:
+			addScenarioNode(sandboxName, nodeName)
+		case mod.EventModifyNode:
```
roymx, line 187: Thought we would not support the Modify operation
How much testing went into modifying a running process?

roymx (approved):

go-apps/meep-sandbox-ctrl/server/sandbox-ctrl.go
dilallkx: The previous implementation supported only UE type. Now the API supports all physical location and process types. This function was updated to check for all physical location and process types.

go-apps/meep-virt-engine/server/virt-engine.go
dilallkx: The frontend currently supports only Add & Remove of processes.
The API supports Add/Remove & Modify of physical locations & processes. 
The Virt Engine supports Add/Remove & Modify of physical locations & processes. Basic manual testing of the Modify was done for now and the Virt Engine successfully redeployed the process. If the TC Engine & MG Manager components cannot easily support these API changes during implementation then we can disable support Modify. For now I would leave it enabled.

PR #214, opened 2021-03-01
Modified couchdb authentication process

nikhildoifode:
- Removed cookie timeout
- Added go routine to authenticate every 9 minutes

### DIFF
--- a/charts/couchdb/values.yaml
+++ b/charts/couchdb/values.yaml
@@ -192,8 +192,8 @@ couchdbConfig:
     # 5984 when is set to true.
     require_valid_user: false
   couch_httpd_auth:
-    # Setting default timeout to 365 days (365*24*60*60)
-    # timeout: 31536000
+    # Setting default timeout to 20 min (20*60)
+    timeout: 1200
     require_valid_user: true
     allow_persistent_cookies: true
 
--- a/charts/couchdb/values.yaml
+++ b/charts/couchdb/values.yaml
@@ -193,7 +193,7 @@ couchdbConfig:
     require_valid_user: false
   couch_httpd_auth:
     # Setting default timeout to 365 days (365*24*60*60)
-    timeout: 31536000
+    # timeout: 31536000
     require_valid_user: true
     allow_persistent_cookies: true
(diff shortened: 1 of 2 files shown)


charts/couchdb/values.yaml
```diff
@@ -193,7 +193,7 @@ couchdbConfig:
     require_valid_user: false
   couch_httpd_auth:
     # Setting default timeout to 365 days (365*24*60*60)
-    timeout: 31536000
+    # timeout: 31536000
```
dilallkx, line 196: We should explicitly set a default value to use so that it is clear what the timeout value is and what the connection authentication timer is. I suggest using 20 min session timeout with a 10 minute session refresh timer. This will match the values used in the meep-session package.

go-packages/meep-couch/db.go
```diff
@@ -56,12 +67,18 @@ func NewConnector(addr string, dbName string) (rc *Connector, err error) {
...
 
-		var address []string = strings.SplitAfter(addr, "http://")
-		var newAddr = address[0] + "admin:admin@" + address[1]
-		c.dbClient, err = kivik.New(context.TODO(), "couch", newAddr)
+		c.dbClient, err = kivik.New(context.TODO(), "couch", addr)
 		if err != nil {
 			return nil, err
 		}
+
+		err = c.dbClient.Authenticate(context.TODO(), &chttp.CookieAuth{Username: "admin", Password: "admin"})
```
dilallkx, line 75: We should pass these credentials using an environment variable instead. Please create a new task for this that we can address later. For now, we can leave it as is.

dilallkx (changes requested):

charts/couchdb/values.yaml
nikhildoifode: Changed the timeout to 20 min and refresh timer to 10 min

go-packages/meep-couch/db.go
nikhildoifode: ok

dilallkx (approved):

PR #220, opened 2021-03-08
App Lifecycle: TC Engine & TC Sidecar support

dilallkx:
**CHANGES:**
- Prometheus Chart:
  - Set default retention period to 30d
- MG Manager:
  - Updated log format to text instead of JSON
  - Improved code by reinitializing HTTP logger only when scenario name changes
- TC Engine:
  - New IP Manager with goroutine to periodically check for IP address updates
  - New Routing Manager to logically separate Routing Rules from Shaping & Filter rules
  - Support for adding & removing physical locations & processes
  - TC Engine threads & event handlers implemented as follows:
    - **Thread 1: _MQ Handler_**
      - Scenario Activated:
        - Process scenario
        - Refresh NC rules
        - Refresh LB rules
        - Start IP Manager
        - Start NC Manager
      - Scenario Terminated:
        - Remove all rules
        - Publish updates to Sidecars
      - Scenario Updated:
        - Process scenario
        - Refresh NC rules
        - Trigger IP address refresh (only for Add/Remove/Modify Scenario)
        - Refresh LB rules (only for Add/Remove/Modify Scenario)
      - MG Svc Map Update:
        - Refresh LB rules
    - **Thread 2: _NC Manager_**
      - Net Char update:
        - Update target NC rule only
      - Update complete:
        - Publish update to Sidecars
    - **Thread 3: _IP Manager_**
      - IP Address update:
        - Update cached IP addresses
        - Refresh NC rules
        - Refresh LB rules
- TC Sidecar:
  - Pinger implementation cleanup
  - Updates to ignore rules where IP addresses are not yet resolved

**TESTING:**
- Cypress & UT pass

### DIFF
--- a/go-apps/meep-tc-engine/ip-mgr.go
+++ b/go-apps/meep-tc-engine/ip-mgr.go
@@ -70,8 +70,6 @@ func (im *IpManager) RefreshPodList(podList map[string]bool) {
 	for podName := range im.podIpMap {
 		if _, found := podList[podName]; !found {
 			delete(im.podIpMap, podName)
-		} else {
-			delete(podList, podName)
 		}
 	}
 
@@ -92,8 +90,6 @@ func (im *IpManager) RefreshSvcList(svcList map[string]bool) {
 	for svcName := range im.svcIpMap {
 		if _, found := svcList[svcName]; !found {
 			delete(im.svcIpMap, svcName)
-		} else {
-			delete(svcList, svcName)
 		}
 	}
 
--- a/go-apps/meep-tc-engine/ip-mgr.go
+++ b/go-apps/meep-tc-engine/ip-mgr.go
@@ -43,7 +43,7 @@ type IpManager struct {
 	clientset   *kubernetes.Clientset
 	updateCb    IpAddrUpdateCb
 	isConnected bool
-	mutex       *sync.Mutex
+	mutex       sync.Mutex
 }
 
 // NewIpManager - Creates and initialize an IP Manager instance
@@ -62,16 +62,43 @@ func NewIpManager(name string, sandboxName string, updateCb IpAddrUpdateCb) (im 
 }
 
 // SetPodList - Set list of pods to monitor IP addresses for
-func (im *IpManager) SetPodList(podList []string) {
+func (im *IpManager) SetPodList(podList map[string]bool) {
 	im.mutex.Lock()
 	defer im.mutex.Unlock()
 
+	// Remove stale entries
+	for podName := range im.podIpMap {
+		if _, found := podList[podName]; !found {
+			delete(im.podIpMap, podName)
+		}
+	}
+
+	// Add missing entries
+	for podName := range podList {
+		if _, found := im.podIpMap[podName]; !found {
+			im.podIpMap[podName] = IP_ADDR_NONE
+		}
+	}
 }
 
 // SetSvcList - Set list of services to monitor IP addresses for
-func (im *IpManager) SetSvcList(svcList []string) {
+func (im *IpManager) SetSvcList(svcList map[string]bool) {
 	im.mutex.Lock()
 	defer im.mutex.Unlock()
+
+	// Remove stale entries
+	for svcName := range im.svcIpMap {
+		if _, found := svcList[svcName]; !found {
+			delete(im.svcIpMap, svcName)
+		}
+	}
+
+	// Add missing entries
+	for svcName := range svcList {
+		if _, found := im.svcIpMap[svcName]; !found {
+			im.svcIpMap[svcName] = IP_ADDR_NONE
+		}
+	}
 }
 
 // GetPodIp - Get Pod IP address
@@ -85,7 +112,7 @@ func (im *IpManager) GetPodIp(podName string) string {
 
 // GetPodIp - Get Svc IP address
 func (im *IpManager) GetSvcIp(svcName string) string {
-	svcIp, found := im.podIpMap[svcName]
+	svcIp, found := im.svcIpMap[svcName]
 	if !found {
 		svcIp = IP_ADDR_NONE
 	}
@@ -171,13 +198,6 @@ func (im *IpManager) refreshIpAddresses() {
 			log.Debug("Setting podName: ", podName, " ip: ", podIp)
 			im.podIpMap[podName] = podIp
 			updated = true
-
-			// //set the element if it has already been created by the scenario parsing
-			// element := netElemMap[podName]
-			// if element != nil {
-			// 	element.Ip = podIp
-			// 	element.NextUniqueNumber = 1
-			// }
 		}
 	}
 
--- a/go-apps/meep-tc-engine/ip-mgr.go
+++ b/go-apps/meep-tc-engine/ip-mgr.go
@@ -61,8 +61,8 @@ func NewIpManager(name string, sandboxName string, updateCb IpAddrUpdateCb) (im 
 	return im, nil
 }
 
-// SetPodList - Set list of pods to monitor IP addresses for
-func (im *IpManager) SetPodList(podList map[string]bool) {
+// RefreshPodList - Set list of pods to monitor IP addresses for
+func (im *IpManager) RefreshPodList(podList map[string]bool) {
 	im.mutex.Lock()
 	defer im.mutex.Unlock()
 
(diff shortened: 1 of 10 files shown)


go-apps/meep-tc-engine/ip-mgr.go
```diff
@@ -0,0 +1,244 @@
...
+	im.sandboxName = sandboxName
+	im.updateCb = updateCb
+	im.podIpMap = make(map[string]string)
+	im.svcIpMap = make(map[string]string)
+
+	log.Info("Successfully create IP Manager")
+	return im, nil
+}
+
+// SetPodList - Set list of pods to monitor IP addresses for
```
roymx, line 64: TBD - comment

go-apps/meep-tc-engine/ip-mgr.go
```diff
@@ -0,0 +1,244 @@
...
+	im.ticker = time.NewTicker(DEFAULT_TICKER_INTERVAL_MS * time.Millisecond)
+	go func() {
+		for range im.ticker.C {
+			im.mutex.Lock()
+
+			// Refresh IP addresses
+			im.refreshIpAddresses()
+
+			im.mutex.Unlock()
+		}
```
roymx, line 144: Should call `Refresh()` directly

go-apps/meep-tc-engine/ip-mgr.go
```diff
@@ -0,0 +1,244 @@
...
+	// Start ticker to periodically retrieve platform information
+	im.ticker = time.NewTicker(DEFAULT_TICKER_INTERVAL_MS * time.Millisecond)
+	go func() {
+		for range im.ticker.C {
+			im.mutex.Lock()
+
+			// Refresh IP addresses
+			im.refreshIpAddresses()
+
+			im.mutex.Unlock()
```
roymx, line 143: Should call `Refresh()` directly

go-apps/meep-tc-engine/ip-mgr.go
```diff
@@ -0,0 +1,244 @@
...
+		im.clientset, err = connectToAPISvr()
+		if err != nil {
+			log.Error("Failed to connect with k8s API Server. Error: ", err)
+			return
+		}
+		im.isConnected = true
+	}
+
+	// Retrieve all sandbox pods from k8s api
+	podList, err := im.clientset.CoreV1().Pods(im.sandboxName).List(metav1.ListOptions{})
```
roymx, line 185: I believe the above query will return all pods in the sandbox (core & scenario)
e.g. `kubectl get pods -n <sandboxName>`

For efficiency, it should only focus at the scenario pods
e.g ` kubectl get pods -n <sandboxName> -l meepOrigin=scenario`

go-apps/meep-tc-engine/ip-mgr.go
```diff
@@ -0,0 +1,244 @@
...
+
+		if ip, found := im.podIpMap[podName]; found && podIp != "" && podIp != ip {
+			log.Debug("Setting podName: ", podName, " ip: ", podIp)
+			im.podIpMap[podName] = podIp
+			updated = true
+		}
+	}
+
+	// Retrieve all sandbox services from k8s api
+	svcList, err := im.clientset.CoreV1().Services(im.sandboxName).List(metav1.ListOptions{})
```
roymx, line 205: I believe the above query will return all pods in the sandbox (core & scenario)
e.g. `kubectl get service -n <sandboxName>`

For efficiency, it should only focus at the scenario pods
e.g ` kubectl get service -n <sandboxName> -l meepOrigin=scenario`

Unfortunately, the services do not have the meepOrigin label - make sure that scenario created services get the same label set as the pods

go-apps/meep-tc-engine/ip-mgr.go
```diff
@@ -0,0 +1,244 @@
...
+// SetPodList - Set list of pods to monitor IP addresses for
+func (im *IpManager) SetPodList(podList map[string]bool) {
+	im.mutex.Lock()
+	defer im.mutex.Unlock()
+
+	// Remove stale entries
+	for podName := range im.podIpMap {
+		if _, found := podList[podName]; !found {
+			delete(im.podIpMap, podName)
+		}
```
roymx, line 73: When the element is found in this iteration it should be removed from `podList` so the second iteration is faster (only contains `!found` things)
else {
    delete(podList, podName)

go-apps/meep-tc-engine/ip-mgr.go
```diff
@@ -0,0 +1,244 @@
...
+// SetSvcList - Set list of services to monitor IP addresses for
+func (im *IpManager) SetSvcList(svcList map[string]bool) {
+	im.mutex.Lock()
+	defer im.mutex.Unlock()
+
+	// Remove stale entries
+	for svcName := range im.svcIpMap {
+		if _, found := svcList[svcName]; !found {
+			delete(im.svcIpMap, svcName)
+		}
```
roymx, line 93: When the element is found in this iteration it should be removed from `svcList` so the second iteration is faster (only contains `!found` things)
else {
    delete(svcList, svcName)

go-apps/meep-tc-engine/ip-mgr.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// Add missing entries
+	for podName := range podList {
+		if _, found := im.podIpMap[podName]; !found {
+			im.podIpMap[podName] = IP_ADDR_NONE
+		}
+	}
+}
+
+// SetSvcList - Set list of services to monitor IP addresses for
```
roymx, line 84: This function performs an "Update" / "Refresh" operation rather than "Set" & should be named accordingly

go-apps/meep-tc-sidecar/main.go
```diff
@@ -249,21 +254,40 @@ func initMeepSidecar() (err error) {
...
+		return err
+	}
+
+	// Create & initialize pinger instance
+	pinger, err = New(opts.bind4, opts.bind6)
+	if err != nil {
+		log.Error("Failed to create Pinger. Error: ", err)
+		return err
+	}
+	if pinger.PayloadSize() != uint16(opts.payloadSize) {
```
roymx, line 267: Why check `pinger.PayloadSize() != uint16(opts.payloadSize)` - just set it.

go-apps/meep-tc-engine/ip-mgr.go
dilallkx: Modified as suggested

go-apps/meep-tc-engine/ip-mgr.go
dilallkx: Modified as suggested

go-apps/meep-tc-engine/ip-mgr.go
dilallkx: Modified as suggested

go-apps/meep-tc-engine/ip-mgr.go
dilallkx: Agreed. Modifying as suggested. Note that the meepOrigin label is inserted in every deployment & statefulset (and hence the pods) using the meep-webhook, which guarantees that the label will be present.

go-apps/meep-tc-engine/ip-mgr.go
dilallkx: The meep-webhook does not inject the meepOrigin label on services, therefore it is not guaranteed that it will be present on all scenario services (e.g. user chart services). The finest granularity for services is then the sandbox/namespace, which is what is used in the code. To avoid modifying the meep-webhook at this point, I recommend keeping the code as is and getting the IP addresses for all sandbox services.

Additional note: meep-webhook requires the user-defined helm charts to set the 'release' label on its deployments, which allows the webhook to identify scenario pods where the meep-tc-sidecar must be injected. Typical helm charts set this label by default on deployments, however they do not set them on services. Adding support for service label injection in the meep-webhook would add a requirement on user-charts to set the 'release' label on all services as well. To reduce complexity and avoid this additional requirement it makes sense to add slightly more processing in the IP Manager code to retrieve IP addresses for all sandbox services instead of only the scenario services.

go-apps/meep-tc-engine/ip-mgr.go
dilallkx: Modified as suggested

go-apps/meep-tc-engine/ip-mgr.go
dilallkx: Modified as suggested

go-apps/meep-tc-sidecar/main.go
dilallkx: Modified as suggested.

roymx (approved):
Thanks for the updates.

PR #222, opened 2021-03-15
App Lifecycle: MG Manager support

dilallkx:
**CHANGES:**
- meep-model package:
  - New method to update POAs in range (previously in Sandbox Controller)
- TC Engine:
  - Include Pod IP in LB rule updates to help sidecar determine when to apply the rules
- TC Sidecar:
  - Throughput calculation improvement (timestamp now taken immediately after retrieving metrics)
  - LB rules check for valid IP address before applying rules
- MG Manager:
  - Remove unused maps
  - Added message-specific handlers (activate, update, terminate scenario)
  - Code updates to remove stale Mobility Groups, MG Apps & UEs
  - New validation checks: 
    - MG App name must be equal to a service instance name in active scenario
    - UE ID must be equal to a UE name in active scenario
  - LB hop count algorithm update to stay on current edge app if hop count is equal

**TESTING:**
- UT, System & Cypress tests pass
- Manual verification of Add/Remove Edge Applications with Multi-Edge services

**KNOWN LIMITATIONS:**
- App state transfer not supported if the MG App instance serving the UE is removed
  - The state cannot be retrieved from the removed MG app instance and therefore cannot be transferred
- App state transfer not supported if the new MG app instance is used immediately
  - The state is passed to the MG Manager, however the MG Manager currently does not store the state if the target app is not immediately reachable
  - NOTE: if the new MG App instance has enough time to start before the app state transfer then it will be successful

### DIFF
--- a/go-apps/meep-mg-manager/server/mg-manager.go
+++ b/go-apps/meep-mg-manager/server/mg-manager.go
@@ -941,6 +941,9 @@ func mgAppCreate(mgName string, mgApp *mgModel.MobilityGroupApp) error {
 	// Store & Apply latest MG Service mappings
 	applyMgSvcMapping()
 
+	// Inform TC Engine of LB rules updatge
+	publishLbRulesUpdate()
+
 	return nil
 }
 
@@ -1038,6 +1041,9 @@ func mgUeCreate(mgName string, appID string, mgUe *mgModel.MobilityGroupUe) erro
 
 		// Store & Apply latest MG Service mappings
 		applyMgSvcMapping()
+
+		// Inform TC Engine of LB rules updatge
+		publishLbRulesUpdate()
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


go-apps/meep-mg-manager/server/mg-manager.go
```diff
@@ -319,12 +367,21 @@ func clearScenario() {
 func processScenario(model *mod.Model) error {
 	log.Debug("processScenario")
 
+	// Reset service info maps
+	mgm.svcInfoMap = make(map[string]*serviceInfo)
+	mgm.mgSvcInfoMap = make(map[string]*mgServiceInfo)
+	mgm.mgList = make(map[string]bool)
```
roymx, line 373: - If the above maps share the same keys - it may be beneficial to maintain a single map.
- `mgList` is a map

roymx (approved):

go-apps/meep-mg-manager/server/mg-manager.go
dilallkx: You are correct. I removed mgList and now use mgSvcInfoMap instead.

PR #241, opened 2021-04-27
Meep-model package improvements for getting active scenario nodes

dilallkx:
**CHANGES:**
- Removed 404 response codes when list is empty (successfully found endpoint but returns an empty list)
- Fixed formatting issues
- Implemented logic as described in the previous review comments

**TESTING:**
- UT & Cypress tests pass

### DIFF
--- a/go-packages/meep-model/model.go
+++ b/go-packages/meep-model/model.go
@@ -19,7 +19,6 @@ package model
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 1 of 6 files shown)


go-packages/meep-model/model.go
```diff
@@ -957,590 +956,211 @@ func (m *Model) GetNodeContext(name string) (ctx interface{}) {
...
-	case Proc:
-		if filter.ProcessName != "" {
-			if filter.ProcessType != "" {
-				n = m.nodeMap.FindByType(filter.ProcessName, filter.ProcessType)
-				if n != nil {
-					nList = append(nList, n)
+
+				// Remove children
+				if !filter.Children {
+					domain.Zones = nil
```
roymx, line 987: The logic `!filter.Children` is awkward
- if `filter.children == false` then childrens are removed
- if `filter.children == true` then childrens are kept

roymx (approved):

go-packages/meep-model/model.go
dilallkx: Agreed. I will update API & backend accordingly.

dilallkx: NOTE: I will push the necessary updates in an upcoming PR.

PR #247, opened 2021-05-10
Power signal updates on active scenario updates

pastorsx:
Power signals in GIScache updated every time a UE changes position.

GIS cache power signal were only red on a 1s refresh ticker in RNIS. The GIS cache info needs to be read whenever an active scenario update is received too, otherwise, measurements information (power signal) is only read after the ticker... which might be too late after a notification with the "signal power" is sent. \

This fix ensure they are read in both instances

test:
UT passed

### DIFF
--- a/go-apps/meep-rnis/sbi/rnis-sbi.go
+++ b/go-apps/meep-rnis/sbi/rnis-sbi.go
@@ -29,17 +29,21 @@ import (
 const moduleName string = "meep-rnis-sbi"
 
 type UeDataSbi struct {
-	Name         string
-	Mnc          string
-	Mcc          string
-	CellId       string
-	NrCellId     string
-	ErabIdValid  bool
-	AppNames     []string
-	Latency      int32
-	ThroughputUL int32
-	ThroughputDL int32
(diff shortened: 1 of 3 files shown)


go-apps/meep-rnis/sbi/rnis-sbi.go
```diff
@@ -326,18 +330,26 @@ func processActiveScenarioUpdate() {
 						throughputUL = ue.NetChar.ThroughputUl
 					}
 
+					//update measurements, don't wait for ticker
+					ueMeasMap, _ := sbi.gisCache.GetAllMeasurements()
```
dilallkx, line 334: We should call this outside the UE iteration loop if there is at least 1 UE.

dilallkx (changes requested):

go-apps/meep-rnis/sbi/rnis-sbi.go
pastorsx: good catch, fixed

dilallkx (approved):

PR #258, opened 2021-05-26
MEC028 (wais) v2.1.10 update and StaDataRate subscription

pastorsx:
MEC028 update with 2.1.10 data structures and endpoints
- compile previous code and report notification as it used to
- need a follow-up update to have the notification matching the new subscription triggers and threshold (subscription behavior changed'
- added staDataRate subscription
  - SBI updates to track apps under a UE
  - update to the code to get the downlink and uplink traffic from the metric store
  - storage of the downlink and uplink metrics
  - notification based on updates to the scenario (same as assocStaNotification). Valid but missing the periodic trigger (to be done as part of the assocSta update - same periodic trigger)

testing
- scenario using wifi POAs, notification generated as expected based on triggers
- ran the current system test (passed).

Code update
- lots of file changed caused by the new yaml file being generated
- real changes limited to :
    go-apps/meep-wais/sbi/wais-sbi.go
    go-apps/meep-wais/server/wais.go
    go-apps/meep-wais/server/wais_test.go
    go-apps/meep-wais/server/convert.go
    go-apps/meep-wais/api/swagger.yaml
    test/system/wais_test.go


go-apps/meep-wais/server/wais.go
```diff
@@ -178,49 +198,76 @@ func Stop() (err error) {
...
+	if jsonStaData != "" {
+		staData = convertJsonToStaData(jsonStaData)
+	}
+
+	//update data rates
+	sumUl := 0.0
+	sumDl := 0.0
+	for _, appName := range appNames {
+
+		metricsArray, err := metricStore.GetCachedNetworkMetrics("*", appName)
```
dilallkx, line 215: This should be done in the SBI.

dilallkx (changes requested):

go-apps/meep-wais/server/wais.go
pastorsx: this is the function updateStaInfo which is called from the sbi but implemented in the wais.go file.... So i think that is the SBI. can you confirm you agree?

dilallkx: Yes, the SBI invokes this function, however the wais now depends on the metric store to retrieve the network metrics. I believe this dependency should be in the SBI package only. The SBI package should be the only one that changes to obtain the necessary data.

pastorsx: moved the metricStore code to the sbi

go-apps/meep-wais/sbi/wais-sbi.go
```diff
@@ -195,8 +223,17 @@ func processActiveScenarioUpdate() {
...
+	sbi.updateScenarioNameCB(scenarioName, &update)
+
+	// Connect to Metric Store
+	if update {
+		var err error
+		metricStore, err = met.NewMetricStore(scenarioName, sbi.sandboxName, influxAddr, redisAddr)
+		if err != nil {
+			log.Error("Failed connection to metric-store: ", err)
+		}
+	}
```
dilallkx, line 236: The SBI should not depend on the WAIS to determine when to connect to the metric store.

dilallkx (changes requested):

dilallkx:
Also, please rebase this branch to the tip of develop.

go-apps/meep-wais/sbi/wais-sbi.go
pastorsx: i put all the logic in sbi now

dilallkx (approved):

PR #259, opened 2021-05-31
new GIS engine API DistanceTo and WithinRange endoints

pastorsx:
Adding 2 new endpoints:
- DistanceTo
- WithinRange

UT test updates

passed UT test


go-apps/meep-gis-engine/api/swagger.yaml
```diff
@@ -283,6 +350,50 @@ definitions:
     example:
       active: true
       type: "MOBILITY"
+  DistanceParameters:
```
dilallkx, line 353: Would it be clearer to call this object _TargetPoint_ and use _TargetRange_ for the withinRange endpoint?

go-apps/meep-gis-engine/api/swagger.yaml
```diff
@@ -283,6 +350,50 @@ definitions:
...
+        type: "number"
+        format: "float"
+        description: "Longitude of a second element for query purpose. Taken into\
+          \ account only if AssetName is not present."
+    description: "Parameters for distance query purpose."
+    example:
+      latitude: 0.8008282
+      assetName: "assetName"
+      longitude: 6.0274563
+  DistanceResponse:
```
dilallkx, line 375: We could call this object simple _Distance_. i.e. it would return a _Distance_ object.

go-apps/meep-gis-engine/api/swagger.yaml
```diff
@@ -283,6 +350,50 @@ definitions:
...
+        format: "float"
+        description: "Distance between two points (in meters)"
+      latitude:
+        type: "number"
+        format: "float"
+        description: "Destination asset latitude"
+      longitude:
+        type: "number"
+        format: "float"
+        description: "Destination asset longitude"
```
dilallkx, line 391: Why return these? Are they the coordinates that were passed in the request?
Should we return the src & dest coordinates if we want to include these values?

go-apps/meep-gis-engine/api/swagger.yaml
```diff
@@ -320,6 +434,62 @@ definitions:
...
+        format: "float"
+        description: "Accuracy of the location (in meters)."
+    description: "Parameters for within range query purpose."
+    example:
+      latitude: 0.8008282
+      assetName: "assetName"
+      accuracy: 5.962134
+      radius: 1.4658129
+      longitude: 6.0274563
+  WithinRangeResponse:
```
dilallkx, line 471: Again, recommend directly returing a _WithinRange_ object here.

go-apps/meep-gis-engine/api/swagger.yaml
```diff
@@ -320,6 +434,62 @@ definitions:
...
+    - "within"
+    properties:
+      latitude:
+        type: "number"
+        format: "float"
+        description: "Destination asset latitude"
+      longitude:
+        type: "number"
+        format: "float"
+        description: "Destination asset longitude"
```
dilallkx, line 483: Again, unclear which point this refers to. Should we include both src & dst + radiues here? or none?

go-apps/meep-gis-engine/server/gis-engine.go
```diff
@@ -1106,6 +1107,252 @@ func geGetAssetData(w http.ResponseWriter, r *http.Request) {
...
+		}
+	} else {
+		dstLong = distanceParam.Longitude
+		dstLat = distanceParam.Latitude
+		dstLongStr = strconv.FormatFloat(float64(distanceParam.Longitude), 'f', -1, 32)
+		dstLatStr = strconv.FormatFloat(float64(distanceParam.Latitude), 'f', -1, 32)
+	}
+
+	srcCoordinates := "(" + srcLongStr + " " + srcLatStr + ")"
+	dstCoordinates := "(" + dstLongStr + " " + dstLatStr + ")"
```
dilallkx, line 1208: Why convert these to strings? Why not leave them as GeoJSON point coordinates.
Perhaps the API should take a point instead of a lat/long as well, this would be in line with the existing GeoData format.

go-packages/meep-gis-asset-mgr/asset-mgr.go
```diff
@@ -2009,3 +2009,72 @@ func calculateWifiPower(radius float32, distance float32) (rssi float32) {
 	}
 	return rssi
 }
+
+// Get distance between 2 coordinates
```
dilallkx, line 2013: Fix indentation in file.

dilallkx (changes requested):

go-apps/meep-gis-engine/api/swagger.yaml
pastorsx: removed the Response part of the object name

go-apps/meep-gis-engine/api/swagger.yaml
pastorsx: removed the Response part of the object name

go-apps/meep-gis-engine/api/swagger.yaml
pastorsx: is meant to be an optional parameter so that the api can return it or not for the destination only, not the source point. Reason behind it is because the source point is a parameter while the response is just related to the destination that was in the body. It is assumed only the destination information is requested since the source should know where it is located.

go-apps/meep-gis-engine/api/swagger.yaml
pastorsx: I changed it because i have no strong opinion.. i like the parameters better but as i said.. no strong opinion...

go-apps/meep-gis-engine/api/swagger.yaml
pastorsx: same explanation as above

go-packages/meep-gis-asset-mgr/asset-mgr.go
pastorsx: fixed

go-apps/meep-gis-engine/server/gis-engine.go
pastorsx: gis-cache in redis is also storing only as long and lat rather than a json.. should all of them be a similar format to prevent conversions? I left it as is for now...

go-apps/meep-gis-engine/api/swagger.yaml
dilallkx: Okay, it makes sense if the target point is an asset name that we may want to return its coordinates as well. If the Target point is a lat/long I assume you don;t return the lat/long, correct?

go-apps/meep-gis-engine/server/gis-engine.go
dilallkx: We can leave as is for now. If we need to edit/add/use this API differently in the future we can adapt the API & microservices accordingly.

go-packages/meep-gis-engine-client/docs/WithinRangeParameters.md
```diff
@@ -0,0 +1,14 @@
+# WithinRangeParameters
```
dilallkx, line 1: Remove the files that are no longer generated.

dilallkx (changes requested):

go-packages/meep-gis-engine-client/docs/WithinRangeParameters.md
pastorsx: removed

go-apps/meep-gis-engine/server/gis-engine.go
pastorsx: ok

dilallkx (approved):

PR #262, opened 2021-06-08
Location service : Area circle and Distance subscription added, Distance endpoint added

pastorsx:
New subscription added
area/circle
distance

New endpoint added:
distance

UT added and system test added
swagger api updated and moved under proper tag (unsupported only has periodic subscription now)

testing:
ut passed
system test passed


go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -172,6 +215,13 @@ func Init() (err error) {
 	}
 	log.Info("SBI Initialized")
 
+	distancePeriodicTicker = time.NewTicker(time.Second)
```
dilallkx, line 218: Move this code inside the Run() function.
Also, stop the ticker in the Stop() function.

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -311,6 +361,287 @@ func registerUser(userAddress string, event []UserEventType, subsIdStr string) {
...
+					terminalLocation.CurrentLocation = &locationInfo
+					retrievalStatus := RETRIEVED
+					terminalLocation.LocationRetrievalStatus = &retrievalStatus
+					terminalLocationList = append(terminalLocationList, terminalLocation)
+				}
+				distanceNotif.TerminalLocation = terminalLocationList
+				distanceNotif.CallbackData = distanceCheck.Subscription.ClientCorrelator
+				var inlineDistanceSubscriptionNotification InlineSubscriptionNotification
+				inlineDistanceSubscriptionNotification.SubscriptionNotification = &distanceNotif
+				sendSubscriptionNotification(distanceCheck.Subscription.CallbackReference.NotifyURL, inlineDistanceSubscriptionNotification)
```
dilallkx, line 485: Should we run this in a goroutine to avoid remote client from blocking Location Service execution?
NOTE: Same applies for existing notifications.

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -311,6 +361,287 @@ func registerUser(userAddress string, event []UserEventType, subsIdStr string) {
...
+			}
+
+			for _, pair := range addressPairs {
+				refAddr := pair.addr1
+				monitoredAddr := pair.addr2
+
+				var distParam gisClient.TargetPoint
+				distParam.AssetName = monitoredAddr
+
+				distResp, _, err := gisAppClient.GeospatialDataApi.GetDistanceGeoDataByName(context.TODO(), refAddr, distParam)
```
dilallkx, line 418: Can this operation take a long time if we have many subscriptions? Does it make sense to block the entire location service while we process this periodic notification?

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -311,6 +361,287 @@ func registerUser(userAddress string, event []UserEventType, subsIdStr string) {
...
+
+				distance := int32(distResp.Distance)
+
+				switch *distanceCheck.Subscription.Criteria {
+				case ALL_WITHIN_DISTANCE:
+					if float32(distance) < distanceCheck.Subscription.Distance {
+						returnAddr[monitoredAddr] = &distResp
+					} else {
+						skipThisSubscription = true
+						break
```
dilallkx, line 432: You are breaking from the switch statement only and not the addrPairs loop. You skip the subscription, however you may also want to stop sending GIS Engine REST API calls for any remaining address pairs for the current subscription.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx, line 218: Is the periodic timer configurable via the API or does it always have to be 1 second.

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -311,6 +361,287 @@ func registerUser(userAddress string, event []UserEventType, subsIdStr string) {
...
+func checkNotificationAreaCircle(addressToCheck string) {
+
+	//only check if there is at least one subscription
+	mutex.Lock()
+	defer mutex.Unlock()
+	//check all that applies
+	for subsId, areaCircleCheck := range areaCircleSubscriptionMap {
+		if areaCircleCheck != nil && areaCircleCheck.Subscription != nil {
+			//decrement the next time to send a message
+			areaCircleCheck.NextTts--
```
dilallkx, line 501: Is Frequency defined in seconds?
Are you assuming that this function will be called once every second and using a counter instead of a ticker?

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -311,6 +361,287 @@ func registerUser(userAddress string, event []UserEventType, subsIdStr string) {
...
+				if addr != addressToCheck {
+					continue
+				}
+				//check if address is already inside the area or not based on the subscription
+				var withinRangeParam gisClient.TargetRange
+				withinRangeParam.Latitude = areaCircleCheck.Subscription.Latitude
+				withinRangeParam.Longitude = areaCircleCheck.Subscription.Longitude
+				withinRangeParam.Radius = areaCircleCheck.Subscription.Radius
+
+				withinRangeResp, _, err := gisAppClient.GeospatialDataApi.GetWithinRangeByName(context.TODO(), addr, withinRangeParam)
```
dilallkx, line 519: Should we be blocking the entire location service while waiting for this response?

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -311,6 +361,287 @@ func registerUser(userAddress string, event []UserEventType, subsIdStr string) {
...
+				terminalLocation.CurrentLocation = &locationInfo
+				retrievalStatus := RETRIEVED
+				terminalLocation.LocationRetrievalStatus = &retrievalStatus
+				terminalLocationList = append(terminalLocationList, terminalLocation)
+
+				areaCircleNotif.TerminalLocation = terminalLocationList
+				areaCircleNotif.CallbackData = areaCircleCheck.Subscription.ClientCorrelator
+				var inlineCircleSubscriptionNotification InlineSubscriptionNotification
+				inlineCircleSubscriptionNotification.SubscriptionNotification = &areaCircleNotif
+				sendSubscriptionNotification(areaCircleCheck.Subscription.CallbackReference.NotifyURL, inlineCircleSubscriptionNotification)
```
dilallkx, line 576: Some comment, should we be blocking the entire location service while waiting for response?

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -1204,49 +1589,599 @@ func zonalTrafficSubGet(w http.ResponseWriter, r *http.Request) {
...
+	areaCircleSub := body.CircleNotificationSubscription
+
+	if areaCircleSub == nil {
+		log.Error("Body not present")
+		http.Error(w, "Body not present", http.StatusBadRequest)
+		return
+	}
+
+	//checking for mandatory properties
+	if areaCircleSub.CallbackReference == nil || areaCircleSub.CallbackReference.NotifyURL == "" {
```
dilallkx, line 1727: This code seems to be duplicated. Can/should we use a function?

dilallkx (changes requested):

go-apps/meep-gis-engine/server/gis-engine.go
```diff
@@ -1299,20 +1297,17 @@ func geGetWithinRangeGeoDataByName(w http.ResponseWriter, r *http.Request) {
...
 
-	srcLongStr := ""
-	srcLatStr := ""
-	if srcAsset.geoData != nil {
-		srcPosition := srcAsset.geoData.position
-		srcPoint := convertJsonToPoint(srcPosition)
-		srcLongStr = strconv.FormatFloat(float64(srcPoint.Coordinates[0]), 'f', -1, 32)
-		srcLatStr = strconv.FormatFloat(float64(srcPoint.Coordinates[1]), 'f', -1, 32)
-	} else {
+	position, err := ge.gisCache.GetPosition("*", assetName)
```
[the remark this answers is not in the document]
pastorsx: it seems to be the location of the latest data for me... would i have been able to get it somewhere else saving that Redis call?

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: i agree with your comment and that applies to every notifications... the reason why we block and wait for the response is to keep the sequences of the messages. Our system is meant to allow people to understand the behaviors of MEC apps and helping them to design those, not to offer a high performance system... but if we want to go to better efficiency... we can certainly use a go routine call for those notifications. TBD

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: the idea is to have a main 1s timer and that each subscription has a counter the gets decremented each time this timer expires... The reason is that every subscription can have its own frequency (in seconds)... so instead of having N timers... i have 1 timer and N counters that i check .. i believe it is better this way because it can end up being really a lot of timer resources otherwise.

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: comments above explain the reason why i did so... the same applies to other subscription driven by a frequency/period...

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: moved

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: agreed.. do you believe a go function should be invoked for each subscription ? adds efficiency for sure

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: intention was to break from the addrPair loop as you are pointing out not only the innerloop.. since no break is needed in golang, i assumed it would break from the surrounding for loop.. but after double checking your comment is right on that matter. 
code updated

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: same comment as above regarding adding go routines... same conclusion should apply

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: the code is unique for each method and each endpoint. Some are similar but each have their differences too. I think this is a clear section at the beginning of each function, so i would it as is... The ones that are very close are the POST and PUT for each endpoints, which look the same except to PUT also checking for ResourceUrl being there. I think duplication is then pretty limited. That being said, your comment is welcome if you feel it adds to clarity to create functions for each endpoint method.

go-apps/meep-gis-engine/server/gis-engine.go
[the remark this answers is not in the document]
dilallkx: Can you get it from the asset map in memory (i.e. ge.assets[])? It looks like you were using it previously but recently converted it to use the GIS cache. If both locations (memory & cache) have the same information, let's minimize the number of calls to redis DB.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: I'm okay with leaving it as is. If we see performance degradations or longer processing delays then we can address.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: Let's leave it as is for now. If we see performance degradations or longer processing delays then we can address.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: The issue here is that this function is called from the SBI callback (updateUserInfo), and not on a 1 second ticker.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: I'm okay with leaving it as is. If we see performance degradations or longer processing delays then we can address.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: I'm okay with leaving it as is. If we see performance degradations or longer processing delays then we can address.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: okay to leave as is, thanks!

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: okay to leave as is, although it would be fine for each subscription to have its own timer.

go-apps/meep-gis-engine/server/gis-engine.go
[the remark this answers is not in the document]
pastorsx: assets only contained initial info and was not updated, that's redisDB call used.. suggesting not looking for performance at this point...

pastorsx (commented):
removed the part that was using the NextTts so that it is purely event based and will update based on comments later on if any or if some details are provided on the spec...

go-apps/meep-gis-engine/server/gis-engine.go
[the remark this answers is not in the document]
dilallkx: Agreed, if it becomes an issue we can improve this portion.

dilallkx (approved):

PR #263, opened 2021-06-10
WAIS (MEC028): Periodic and Trigger combinations for AssocSta and StaDataRate notifications

pastorsx:
Added the portion of the notificationPeriod support for AssocSta and StaDataRate subscriptions
Updated the system and UT test for WAIS (testing all triggers and combinations)

-also updated gis-engine UT and RNIS system test related to a previous merge that broke those features

test done
UT passed
system test DONE
manual testing done


go-apps/meep-wais/server/wais.go
```diff
@@ -357,7 +375,86 @@ func updateApInfo(name string, apMacId string, longitude *float32, latitude *flo
...
+	//decrease all subscriptions
+	//check all that applies
+	for _, subInfo := range staDataRateSubscriptionInfoMap {
+		if subInfo != nil {
+			subInfo.NextTts--
+			//if no periodic check is needed, value is negative
+			if subInfo.NextTts != 0 {
+				continue
+			} else { //restart the nextTts and continue processing to send notification or not
+				subInfo.NextTts = subInfo.Subscription.NotificationPeriod
```
dilallkx, line 440: The code seems to do the right job for triggering periodic notifications, however it is difficult to follow/debug in some places. For example, subInfo.NextTts will cause a notification to be sent when it is set to max value. It may be clearer to add another flag in subInfo to indicate that period has expired in this case and that a notification may be sent if necessary.

go-apps/meep-wais/server/wais.go
```diff
@@ -513,6 +622,19 @@ func checkAssocStaNotificationRegisteredSubscriptions(staMacIds []string, apMacI
...
 					default:
 					}
+
+					//if the notification already triggered, do not send it again unless its a periodic event
+					if match {
+						if !isPeriodicInvoked {
+							if !subInfo.Triggered {
+								assocStaSubscriptionInfoMap[subsId].Triggered = true
+							} else {
+								match = false
```
dilallkx, line 632: Same comment about debugging. In this case "match" is set to false even when there is a match, correct? It may be clearer to use another flag here to indicate that periodic is enabled but should not be triggered until the timer expires.

dilallkx (changes requested):

go-apps/meep-wais/server/wais.go
```diff
@@ -546,18 +668,30 @@ func checkAssocStaNotificationRegisteredSubscriptions(staMacIds []string, apMacI
...
-	mutex.Lock()
-	defer mutex.Unlock()
+	if needMutex {
+		mutex.Lock()
+		defer mutex.Unlock()
+	}
 	//check all that applies
-	for subsId, sub := range staDataRateSubscriptionMap {
+	for subsId, subInfo := range staDataRateSubscriptionInfoMap {
+		log.Info("SIMON check")
```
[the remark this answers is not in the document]
pastorsx: removed

go-apps/meep-wais/server/wais.go
```diff
@@ -589,6 +723,8 @@ func checkStaDataRateNotificationRegisteredSubscriptions(staId *StaIdentity, dat
 				}
 				if match {
 					if sub.NotificationEvent != nil {
+						log.Info("SIMON event DL ", dataRateDl, "---", sub.NotificationEvent.DownlinkRateThreshold)
```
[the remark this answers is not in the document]
pastorsx: removed

go-apps/meep-wais/server/wais.go
pastorsx: added a flag to make it clearer... but essentially same logic

go-apps/meep-wais/server/wais.go
pastorsx: changed to give it a clearer name using another flag within the for loop.. but essentially same logic

dilallkx (approved):

PR #271, opened 2021-06-28
AppSupport application termination subscription support

pastorsx:
Implementation of :
GET /app-support/{appInstanceId}/subscriptions
POST /app-support/{appInstanceId}/subscriptions

GET /app-support/{appInstanceId}/subscriptions/{subscriptionId}
PUT /app-support/{appInstanceId}/subscriptions/{subscriptionId}
DELETE /app-support/{appInstanceId}/subscriptions/{subscriptionId}

Note : if not subscribed, prevent the termination confirmation message from being processed
Note2: Not rebased on develop, branched off sp_dev_mec011_phase1 for now, will be renased when tip of develop is up-to-date.

### DIFF
--- a/go-apps/meep-app-enablement/server/app-support/app-support.go
+++ b/go-apps/meep-app-enablement/server/app-support/app-support.go
@@ -38,7 +38,7 @@ import (
 )
 
 const mappsupportBasePath = "/mec_app_support/v1/"
-const mappsupportKey = "as-sub"
+const mappsupportKey = "as"
 const appEnablementKey = "app-enablement"
 const ACTIVE = "ACTIVE"
 const INACTIVE = "INACTIVE"
@@ -145,14 +145,13 @@ func Init(globalMutex *sync.Mutex) (err error) {
 }
 
 // reInit - finds the value already in the DB to repopulate local stored info
+// NOTE: Init is flushing everything so this is a non-operation code, but if a sbi is added that tracks Activation/Termination of scenarios, then this should become handy, leaving it there for future code updates if needed
 func reInit() {
 	//next available subsId will be overrriden if subscriptions already existed
 	nextSubscriptionIdAvailable = 1
 
-	baseKey := appEnablementBaseKey + ":apps:*:" + mappsupportKey
-	keyName := baseKey + ":subscriptions:" + "*"
+	keyName := appEnablementBaseKey + ":apps:*:" + mappsupportKey + ":subscriptions:" + "*"
 	_ = rc.ForEachJSONEntry(keyName, repopulateAppTerminationNotificationSubscriptionMap, nil)
-
 }
 
 // Run - Start APP support
@@ -165,6 +164,7 @@ func Stop() (err error) {
 	return nil
 }
 
+// see NOTE from ReInit()
 func repopulateAppTerminationNotificationSubscriptionMap(key string, jsonInfo string, userData interface{}) error {
 
 	var subscription AppTerminationNotificationSubscription
--- a/go-apps/meep-app-enablement/server/app-support/app-support.go
+++ b/go-apps/meep-app-enablement/server/app-support/app-support.go
@@ -119,8 +119,8 @@ func Init(globalMutex *sync.Mutex) (err error) {
 	// Set base path
 	basePath = "/" + sandboxName + mappsupportBasePath
 	// Get base store key
-	baseKey = dkm.GetKeyRoot(sandboxName) + mappsupportKey
-	appEnablementBaseKey = dkm.GetKeyRoot(sandboxName) + selfName + ":" + appEnablementKey
+	appEnablementBaseKey = dkm.GetKeyRoot(sandboxName) + appEnablementKey + ":meep:" + selfName
+	baseKey = appEnablementBaseKey + ":" + mappsupportKey
 
 	// Connect to Redis DB
 	rc, err = redis.NewConnector(redisAddr, APP_ENABLEMENT_DB)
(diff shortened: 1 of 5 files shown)


go-apps/meep-app-enablement/server/app-support/app-support.go
```diff
@@ -36,10 +38,11 @@ import (
 )
 
 const mappsupportBasePath = "/mec_app_support/v1/"
-const mappsupportKey = "mec-app-support"
+const mappsupportKey = "as-sub"
```
dilallkx, line 41: The name is slightly misleading here. "as-sub" refers to the subscriptions key rather than a mec-app-support base key. Suggest renaming variable to "subscriptionsKey".

go-apps/meep-app-enablement/server/app-support/app-support.go
```diff
@@ -144,6 +148,11 @@ func Init(globalMutex *sync.Mutex) (err error) {
 func reInit() {
 	//next available subsId will be overrriden if subscriptions already existed
 	nextSubscriptionIdAvailable = 1
+
+	baseKey := appEnablementBaseKey + ":apps:*:" + mappsupportKey
+	keyName := baseKey + ":subscriptions:" + "*"
```
dilallkx, line 153: No need for "subscriptions" level in key. Simply use this key:
`keyName := appEnablementBaseKey + ":apps:*:" + subscriptionsKey + "*"`

go-apps/meep-app-enablement/server/app-support/app-support.go
```diff
@@ -156,6 +165,33 @@ func Stop() (err error) {
 	return nil
 }
 
+func repopulateAppTerminationNotificationSubscriptionMap(key string, jsonInfo string, userData interface{}) error {
```
dilallkx, line 168: Is this code required? Was it meant to handle cases where the pod restarts?
NOTE: on a pod restart the init function will flush the "app-enablement" redis entries, including the subscriptions. Is there any case where this code will be invoked on a list that is not empty?

go-apps/meep-app-enablement/server/app-support/app-support.go
```diff
@@ -231,6 +267,22 @@ func applicationsConfirmTerminationPOST(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	//look if subscription exist to process the Termination POST
+	found := false
+	//loop through appTerm map
+	for _, appTermSubscription := range appTerminationNotificationSubscriptionMap {
```
dilallkx, line 273: Why do you need this map? Why not just get the subscriptions for the specific app instance ID from redis?

go-apps/meep-app-enablement/server/app-support/app-support.go
```diff
@@ -316,3 +369,206 @@ func populateServiceInfoList(key string, jsonInfo string, sInfoList interface{})
...
+	subsIdStr := strconv.Itoa(newSubsId)
+
+	link := new(Self)
+	self := new(LinkType)
+	self.Href = hostUrl.String() + basePath + "applications/" + appInstanceId + "/subscriptions/" + subsIdStr
+	link.Self = self
+	subscription.Links = link
+
+	//registration
+	registerAppTerm(&subscription, newSubsId)
```
dilallkx, line 434: Why register the subscription here? Don't we have the necessary info in the subscription we are storing in redis?

go-apps/meep-app-enablement/server/service-mgmt/service-mgmt.go
```diff
@@ -125,7 +125,7 @@ func Init(globalMutex *sync.Mutex) (err error) {
 	// Set base path
 	basePath = "/" + sandboxName + msmgmtBasePath
 	// Get base store key
-	appEnablementBaseKey = dkm.GetKeyRoot(sandboxName) + selfName + ":" + appEnablementKey
+	appEnablementBaseKey = dkm.GetKeyRoot(sandboxName) + appEnablementKey + ":meep:" + selfName
```
dilallkx, line 128: "meep" should be "mep"

dilallkx (changes requested):

go-apps/meep-app-enablement/server/app-support/app-support.go
pastorsx: changed the key to be "as" only if other resources are added in the future.
For now, only subscriptions can be created but keeping it so it is more generic

go-apps/meep-app-enablement/server/app-support/app-support.go
pastorsx: used on one line

go-apps/meep-app-enablement/server/app-support/app-support.go
pastorsx: added a note that explains ReInit not doing anything irght now because of Flush in Init(), but if sbi is added, would become meaningful

go-apps/meep-app-enablement/server/app-support/app-support.go
pastorsx: maps are used by services to prevent extra calls to the DB... expected performance gains...

go-apps/meep-app-enablement/server/app-support/app-support.go
pastorsx: same as previous comment... it populates the map to minimize queries to the DB (although same info is present in both)

go-apps/meep-app-enablement/server/service-mgmt/service-mgmt.go
pastorsx: fixed

dilallkx (approved):

PR #276, opened 2021-07-15
MEP Support

dilallkx:
**CHANGES:**
- MEC Service helm charts:
  - Added support for multiple instances (i.e. deployment as a MEP scenario pod)
  - Removed hardcoded service names & ingress rules
- App enablement:
  - Removed hardcoded MEP name
  - Modified basepath & keys to support data locality
- Location Service:
  - SBI: support for locality
- MEC Services:
  - Multiple instance support & data isolation
  - New environment variables for app enablement:
    - MEEP_APP_ENABLEMENT: App enablement service name
    - MEEP_SCOPE_OF_LOCALITY: Scope of locality (MEC_SYSTEM or MEC_HOST)
    - MEEP_CONSUMED_LOCAL_ONLY: Can be consumed only local to MEP?
    - MEEP_LOCALITY: Service locality
- Virt Engine:
  - Support for MEP Service deployments (scenario pods deployed using sandbox charts)
  - Patch to retrieve MEP service name from Process image name
  - Sandbox template updates to configure MEP services
- Webhook:
  - Patch to add env var to scenario pods
    - MEEP_POD_NAME
    - MEEP_SANDBOX_NAME
    - MEEP_SCENARIO_NAME
- Model package:
  - Updated context return value + adapted code accordingly

**TESTING:**
- UT, system & cypress tests pass

### DIFF
--- a/go-apps/meep-loc-serv/server/loc-serv.go
+++ b/go-apps/meep-loc-serv/server/loc-serv.go
@@ -207,6 +207,7 @@ func Init() (err error) {
 	log.Info("MEEP_MEP_NAME: ", mepName)
 
 	// Get App Enablement URL
+	appEnablementEnabled = false
 	appEnablementEnv := strings.TrimSpace(os.Getenv("MEEP_APP_ENABLEMENT"))
 	if appEnablementEnv != "" {
 		appEnablementUrl = "http://" + appEnablementEnv
--- a/go-apps/meep-loc-serv/server/loc-serv.go
+++ b/go-apps/meep-loc-serv/server/loc-serv.go
@@ -155,7 +155,7 @@ var sandboxName string
 var mepName string = defaultMepName
 var scopeOfLocality string = defaultScopeOfLocality
 var consumedLocalOnly bool = defaultConsumedLocalOnly
-var locality []string = []string{}
+var locality []string
 var basePath string
 var baseKey string
 var mutex sync.Mutex
@@ -221,7 +221,7 @@ func Init() (err error) {
 	// Get local consumption
 	consumedLocalOnlyEnv := strings.TrimSpace(os.Getenv("MEEP_CONSUMED_LOCAL_ONLY"))
 	if consumedLocalOnlyEnv != "" {
-		value, err := strconv.ParseBool("true")
+		value, err := strconv.ParseBool(consumedLocalOnlyEnv)
 		if err == nil {
 			consumedLocalOnly = value
 		}
@@ -274,6 +274,7 @@ func Init() (err error) {
 	sbiCfg := sbi.SbiCfg{
 		SandboxName:    sandboxName,
 		RedisAddr:      redisAddr,
+		Locality:       locality,
 		UserInfoCb:     updateUserInfo,
 		ZoneInfoCb:     updateZoneInfo,
 		ApInfoCb:       updateAccessPointInfo,
--- a/go-apps/meep-loc-serv/server/loc-serv.go
+++ b/go-apps/meep-loc-serv/server/loc-serv.go
@@ -103,13 +103,9 @@ var userSubscriptionTransferringMap = map[int]string{}
 var userSubscriptionMap = map[int]string{}
 
 var zoneStatusSubscriptionMap = map[int]*ZoneStatusCheck{}
-
 var distanceSubscriptionMap = map[int]*DistanceCheck{}
-
 var periodicTicker *time.Ticker
-
 var areaCircleSubscriptionMap = map[int]*AreaCircleCheck{}
-
 var periodicSubscriptionMap = map[int]*PeriodicCheck{}
 
 var addressConnectedMap = map[string]bool{}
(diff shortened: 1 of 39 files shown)


go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -292,44 +334,74 @@ func Run() (err error) {
...
 }
 
-func getAppInstanceId(appName string, appVersion string) string {
-	//var client *appInfoClient.APIClient
-	appInfoClientCfg := appInfoClient.NewConfiguration()
-	appInfoClientCfg.BasePath = appEnablementClientUrl + "/app_info/v1"
+func startRegistrationTicker() {
+	// Wait a few seconds to allow App Enablement Service to start
+	log.Info("Waiting for App Enablement Service to start")
+	time.Sleep(5 * time.Second)
```
roymx, line 349: Don't like this plain sleep.
Prefer the approach where a service tries to access & sleeps if there is a failure.
Sleep should be 1sec or less.

go-apps/meep-rnis/server/rnis.go
```diff
@@ -189,19 +196,17 @@ type PlmnInfoResp struct {
...
-//MEC011 section begin
 const serviceAppName = "RNI"
 const serviceAppVersion = "2.1.1"
 
 var serviceAppInstanceId string
 
-var appEnablementClientUrl string = "http://meep-app-enablement"
-var appEnablementSupport bool = true
+var appEnablementUrl string
+var appEnablementEnabled bool = false
```
roymx, line 205: From now on, default value for app-enablement should be always enabled...

roymx, line 205: Please don't initialize `appEnablementEnabled` as it suggest that this is the default mode which is not true.

The default mode is set by the environment variable.

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -148,37 +147,33 @@ var influxAddr string = "http://meep-influxdb.default.svc.cluster.local:8086"
...
-//MEC011 section begin
 const serviceAppName = "Location"
 const serviceAppVersion = "2.1.1"
 
 var serviceAppInstanceId string
 
-var appEnablementClientUrl string = "http://meep-app-enablement"
-var appEnablementSupport bool = true
+var appEnablementUrl string
+var appEnablementEnabled bool = false
```
roymx, line 167: Please don't initialize appEnablementEnabled as it suggest that this is the default mode which is not true.

The default mode is set by the environment variable.

go-apps/meep-rnis/server/rnis.go
```diff
@@ -357,54 +401,85 @@ func reInit() {
...
 }
 
-func getAppInstanceId(appName string, appVersion string) string {
-	//var client *appInfoClient.APIClient
-	appInfoClientCfg := appInfoClient.NewConfiguration()
-	appInfoClientCfg.BasePath = appEnablementClientUrl + "/app_info/v1"
+func startRegistrationTicker() {
+	// Wait a few seconds to allow App Enablement Service to start
+	log.Info("Waiting for App Enablement Service to start")
+	time.Sleep(5 * time.Second)
```
roymx, line 427: Same comment as Location Service for the 5sec sleep

go-apps/meep-virt-engine/server/chart-template.go
```diff
@@ -149,6 +154,31 @@ func Deploy(sandboxName string, procName string, model *mod.Model) error {
 	return nil
 }
 
+func getMepService(proc *dataModel.Process) string {
+	// !!! Temporary patch for MEP Service configuration !!!
```
roymx, line 158: Is it still a temporary patch?

roymx (approved):

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: Agreed. The problem is that the initial request usually fails with a timeout that is longer (~20 seconds). To avoid pod startup timing issues a sleep was added here. It doesn't eliminate the possible timeout issue (e.g. if the app enablement service pod takes more than 5 seconds to start up), however it usually prevents the initial request from failing & timing out. We could address this by:
1- making sure Location Service only starts after App Enablement service is running (note: currently can't do this with our scenario pods configuration)
2- fail/timeout more quickly on initial request if App Enablement service is not available (i.e. reduce current ~20 second timeout).
3- Keep a sleep in the MEC Services to allow App enablement service to come up (note: does not guarantee App enablement will be running after sleep, in which case we get a ~20 second timeout anyways)

go-apps/meep-rnis/server/rnis.go
dilallkx: Okay, I will initialize it in the Init() function where the environment variable is read instead.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: Okay, I will initialize it in the Init() function where the environment variable is read instead.

go-apps/meep-rnis/server/rnis.go
dilallkx: Same response.

go-apps/meep-virt-engine/server/chart-template.go
dilallkx: It is the selected implementation for now that could/should eventually change. However, I agree that the comment is misleading here. I will remove it.

go-apps/meep-loc-serv/server/loc-serv.go
dilallkx: After discussion, adding comment in code to clarify the reason for the 5 second sleep.

PR #287, opened 2021-08-02
Termination Notification Support for MEC services

pastorsx:
- added new endpoint to receive notifications from mec011 (only appTermination for now) for:
  - location
  - wai
  - rni
 
No new UT or system test added

testing:
UT and system test passed


go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -3810,3 +3834,67 @@ func distanceGet(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	if !appEnablementEnabled {
+		//just ignore the message
+		w.WriteHeader(http.StatusNoContent)
+		return
+	}
+
+	stopRegistrationTicker()
```
dilallkx, line 3856: Is this call necessary? As soon as we are registered we stop this timer.

go-apps/meep-loc-serv/server/loc-serv.go
```diff
@@ -3810,3 +3834,67 @@ func distanceGet(w http.ResponseWriter, r *http.Request) {
...
+
+	node.Type_ = "EDGE-APP"
+	node.Parent = mepName
+	node.NodeDataUnion = &nodeDataUnion
+
+	eventScenarioUpdate.Node = &node
+	eventScenarioUpdate.Action = "REMOVE"
+
+	event.EventScenarioUpdate = &eventScenarioUpdate
+	event.Type_ = "SCENARIO-UPDATE"
```
dilallkx, line 3883: We should deregister the service & send the scenario update event within the go routine. This will allow us to return immediately and unblock the App enablement microservice.

go-apps/meep-wais/server/wais.go
```diff
@@ -516,18 +526,94 @@ func sendTerminationConfirmation(appInstanceId string) error {
...
+	event.Type_ = "SCENARIO-UPDATE"
+
+	go func() {
+		_, err := sbxCtrlClient.EventsApi.SendEvent(context.TODO(), event.Type_, event)
+		if err != nil {
+			log.Error(err)
+		}
+	}()
+
+	if sendAppTerminationWhenDone {
```
dilallkx, line 608: Same comments as for location service.

dilallkx (changes requested):

go-apps/meep-loc-serv/server/loc-serv.go
pastorsx: overthinking: if registration was not complete... ticker is still running... but then again... service will not receive an appTermination.. so removing this code

dilallkx (approved):

PR #292, opened 2021-08-23
MEC011 system test and sandbox frontend alpha test fixes

pastorsx:
AdvantEDGE fixes resulting from system test and changes needed after sandbox alpha testing


go-apps/meep-app-enablement/server/app-support/app-support.go
```diff
@@ -129,6 +129,9 @@ func Run() (err error) {
 
 // Stop - Stop APP support
 func Stop() (err error) {
+	// Flush all app-enablement instance data
```
dilallkx, line 132: Is there a risk that _basekey_ is not initialized here and we flush the entire redis DB?
We should add a running flag to make sure we are running before stopping the microservice.

go-apps/meep-app-enablement/server/service-mgmt/service-mgmt.go
```diff
@@ -147,6 +147,9 @@ func Run() (err error) {
 
 // Stop - Stop Service Mgmt
 func Stop() (err error) {
+	// Flush all app-enablement instance data
```
dilallkx, line 150: Is there a risk that _basekey_ is not initialized here and we flush the entire redis DB?
We should add a running flag to make sure we are running before stopping the microservice.

test/system/app-enablement_test.go
```diff
@@ -0,0 +1,349 @@
+/*
+ * Copyright (c) 2020  InterDigital Communications, Inc
```
dilallkx, line 2: Should be 2021 on new files.

test/system/app-enablement_test.go
```diff
@@ -0,0 +1,349 @@
...
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"context"
+	"time"
+
+	log "github.com/InterDigitalInc/AdvantEDGE/go-packages/meep-logger"
+	smc "github.com/InterDigitalInc/AdvantEDGE/go-packages/meep-service-mgmt-client"
+        asc "github.com/InterDigitalInc/AdvantEDGE/go-packages/meep-app-support-client"
```
dilallkx, line 29: Fix indentation & formatting for all files.

dilallkx (changes requested):

go-apps/meep-app-enablement/server/app-support/app-support.go
pastorsx: baseKey is set in the init, so i believe it should have a value that is not "".... but i added a check just in case...

go-apps/meep-app-enablement/server/service-mgmt/service-mgmt.go
pastorsx: baseKey is set in the init, so i believe it should have a value that is not "".... but i added a check just in case...

test/system/app-enablement_test.go
pastorsx: fixed

test/system/app-enablement_test.go
pastorsx: didn't do a goimport on that new file.. now i did...

dilallkx (approved):

PR #305, opened 2021-09-16
Frontend client to sandbox controller API Applications & frontend UI

supermikii:
# changes 
* modify map compute pop up to show MEC applications registered
* added redux state handling for mec app instances data
* reduced client fetch & map refresh

### DIFF
--- a/js-apps/meep-frontend/src/js/containers/idc-map.js
+++ b/js-apps/meep-frontend/src/js/containers/idc-map.js
@@ -661,9 +661,9 @@ class IDCMap extends Component {
 
   // UE Marker Event Handler
   updateComputePopup(marker) {
-    if (marker) {
+    var table = this.getTable();
+    if (marker && table && table.entries) {
       // Retrieve state 
-      var table = this.getTable();
       const networkName = getElemFieldVal(table.entries[marker.options.meep.compute.id], FIELD_DN_NAME);
       const edgeProvider = getElemFieldVal(table.entries[marker.options.meep.compute.id], FIELD_DN_ECSP);
       const ladn = getElemFieldVal(table.entries[marker.options.meep.compute.id], FIELD_DN_LADN);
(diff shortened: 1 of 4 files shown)


js-apps/meep-frontend/src/js/containers/idc-map.js
```diff
@@ -661,17 +662,23 @@ class IDCMap extends Component {
   // UE Marker Event Handler
   updateComputePopup(marker) {
     if (marker) {
-      var latlng = marker.getLatLng();
+      // Retrieve state 
       var table = this.getTable();
```
dilallkx, line 666: Need a null check on table. Suggest code like in the UE popup like this:
```
    var table = this.getTable();
    if (marker && table && table.entries) {
```

js-apps/meep-frontend/src/js/containers/meep-container.js
```diff
@@ -588,25 +590,53 @@ class MeepContainer extends Component {
...
+    var assetMap = {
+      ueList: _.sortBy(ueList, ['assetName']),
+      poaList: _.sortBy(poaList, ['assetName']),
+      computeList: _.sortBy(computeList, ['assetName'])
+    };
+
+    //Update UE LIST, COMPUTE LIST, POA list
+    this.props.execChangeMapUeList(assetMap.ueList? _.sortBy(assetMap.ueList, ['assetName']) : []);
+    this.props.execChangeMapPoaList(assetMap.poaList? _.sortBy(assetMap.poaList, ['assetName']) : []);
+    this.props.execChangeMapComputeList(assetMap.computeList? _.sortBy(assetMap.computeList, ['assetName']) : []);
```
dilallkx, line 633: Replace this with a single call to execChangeMap() which is already available in map reducer.

dilallkx (changes requested):

js-apps/meep-frontend/src/js/containers/idc-map.js
supermikii: Thanks, new commit to reflect update

dilallkx (approved):

PR #309, opened 2021-09-29
mec021 ams alpha test fixes

pastorsx:
small fixes including:
- registration should fail when consumer app id does not exist
- nil registration info returned when no registration is stored rather than code 404
- deregister task changed to unsupported tag


go-apps/meep-ams/server/ams.go
```diff
@@ -2001,3 +2022,22 @@ func populateAppInstanceIds(key string, fields map[string]string, response inter
 	//response = &resp
 	return nil
 }
+
+func validateAppInstanceId(appInstanceId string) (error, int, string) {
```
dilallkx, line 2026: You already have a list of App Instance IDs that you obtained from periodic polling the Sandbox Controller applications endpoint. Simply check that the app instance exists in the map instead of looking in redis database.

dilallkx (changes requested):

go-apps/meep-ams/server/ams.go
dilallkx: Fixed.

dilallkx (approved):

PR #318, opened 2021-10-28
Subscriptions package + Websocket support

dilallkx:
**CHANGES:**
- New _meep-subscriptions_ package:
  - Subscriptions management package
  - Uses gorilla/websocket
  - Supports direct http & websocket transports
  - Manages periodic & expiry notifications
  - Unit tests
- _meep-wais_:
  - Integration of _meep-subscriptions_ to manage WAIS subscriptions
- Permissions yaml updates to allow websocket endpoint

**TESTING:**
- UT, Cypress & system tests pass


go-packages/meep-subscriptions/websocket.go
```diff
@@ -0,0 +1,190 @@
...
+	var ws Websocket
+	ws.Cfg = cfg
+
+	// Generate a random websocket URI
+	randomStr, err := generateRand(12)
+	if err != nil {
+		log.Error(err.Error())
+		return nil, err
+	}
+	ws.Id = randomStr
```
roymx, line 67: For the sake of clarity, I would pre-pend the websocket endpoint with `websocket-` (ex: `websocket-12345abcd`)
This way, it would be clearer what this id is (ex. when shown in the MEC Sandbox console)

go-packages/meep-subscriptions/websocket.go
```diff
@@ -0,0 +1,190 @@
...
+	ws.State = WsStateInit
+	ws.Endpoint = "ws/" + randomStr
```
roymx, line 69: Use `ws.Id` instead of `randomStr` here

roymx (changes requested):

go-packages/meep-subscriptions/websocket.go
dilallkx: Fixed.

charts/meep-wais/values-template.yaml
```diff
@@ -55,6 +55,8 @@ ingress:
         {{- end }}
   annotations:
     kubernetes.io/ingress.class: nginx
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "3600"
```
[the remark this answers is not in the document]
dilallkx: This is required to prevent websocket connection timeouts when there are no notifications being sent. This value must be greater than the websocket keepalive time that is currently set to 10 min.

PR #365, opened 2022-07-08
Demo3 Bug Fix

dhruvigoyal25:
This PR makes the following changes to AdvantEDGE:
1. Patch to demo3 service:
    - Updates AMS with contextTransferState when UE moves from one mep zone to another
2. Patch to AMS:
    - If the contextTransferState is complete, AMS to not send a state transfer notification.


dilallkx:
Updates included in PR #393 on mec030 feature branch.

PR #411, opened 2022-12-02
Updated Aggregated Codecov reports script

dhruvigoyal25:
-Added a fix to create a /test/codecov folder if doesn't exist to store the aggregated codecov output files.

### DIFF
--- a/go-apps/meepctl/cmd/test.go
+++ b/go-apps/meepctl/cmd/test.go
@@ -44,6 +44,10 @@ var testCmd = &cobra.Command{
 			return
 		}
 
+		gitDir := viper.GetString("meep.gitdir")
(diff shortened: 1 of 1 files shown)


go-apps/meepctl/cmd/test.go
```diff
@@ -44,6 +45,11 @@ var testCmd = &cobra.Command{
 			return
 		}
 
+		gitDir := viper.GetString("meep.gitdir")
+		if err := os.Mkdir(gitDir+"/test/codecov", os.ModePerm); err != nil {
+			log.Fatal(err)
```
dilallkx, line 50: Why add a new dependency here?
Use fmt to print error and return.

dilallkx (changes requested):

dilallkx (approved):

